PR #6, opened 2022-09-27
Updating Jenkins lib version from previously 20211123 to main

jordarlu:
Signed-off-by: Jeff Lu <chunglu@amazon.com>

### Description
Updating Jenkins shared lib version from previously 20211123 to main 
Updating jobs lib version from previously 20211123 to 1.0.0
Updating existing regression files accordingly 

some comments are at https://github.com/opensearch-project/opensearch-build-libraries/issues/7#issuecomment-1259999902

### Issues Resolved
This update is to match the lib version with the current tag 1.0.0


peterzhuamazon (approved):

tests/jenkins/jobs/CreateGithubIssueExisting_Jenkinsfile.txt
```diff
@@ -5,4 +5,4 @@
             CreateGithubIssueExisting_Jenkinsfile.script(groovy.lang.Closure)
                CreateGithubIssueExisting_Jenkinsfile.createGithubIssue({message=[Error building OpenSearch, retry with: ./build.sh manifests/2.2.0/opensearch-2.2.0.yml --component OpenSearch --snapshot]})
                   createGithubIssue.readYaml({file=manifests/2.0.0/opensearch-2.0.0.yml})
-               CreateGithubIssueExisting_Jenkinsfile.echo(Exception occurred: java.io.FileNotFoundException: manifests/2.0.0/opensearch-2.0.0.yml (No such file or directory))
+               CreateGithubIssueExisting_Jenkinsfile.echo(Exception occurred: java.io.FileNotFoundException: manifests\2.0.0\opensearch-2.0.0.yml (The system cannot find the path specified))
```
gaiksaya, line 8: Looks like this is windows/linux path problem. Can you keep it default to linux for now?

tests/jenkins/jobs/CreateGithubIssue_Jenkinsfile.txt
```diff
@@ -5,4 +5,4 @@
             CreateGithubIssue_Jenkinsfile.script(groovy.lang.Closure)
                CreateGithubIssue_Jenkinsfile.createGithubIssue({message=[Error building OpenSearch, retry with: ./build.sh manifests/2.2.0/opensearch-2.2.0.yml --component OpenSearch --snapshot]})
                   createGithubIssue.readYaml({file=manifests/2.0.0/opensearch-2.0.0.yml})
-               CreateGithubIssue_Jenkinsfile.echo(Exception occurred: java.io.FileNotFoundException: manifests/2.0.0/opensearch-2.0.0.yml (No such file or directory))
+               CreateGithubIssue_Jenkinsfile.echo(Exception occurred: java.io.FileNotFoundException: manifests\2.0.0\opensearch-2.0.0.yml (The system cannot find the path specified))
```
gaiksaya, line 8: Same as above

gaiksaya (changes requested):

dblock:
LGTM but this is going to get annoying quickly. I proposed https://github.com/opensearch-project/opensearch-build-libraries/issues/7 to avoid having to do this every time we release.

gaiksaya:
> LGTM but this is going to get annoying quickly. I proposed #7 to avoid having to do this every time we release.

Hi @dblock , the idea here is each file can have it own version that the file needs to be used. For example, we add some library that will only be used by dashboards. Now only dashboards jenkinsFile will use the new version (1.2.0).
In this way not every jenkinsFile needs to change the version but only the concerned one.

vars/uploadArtifacts.groovy
```diff
@@ -6,7 +6,7 @@
  * compatible open source license.
  */
 void call(Map args = [:]) {
-    def lib = library(identifier: 'jenkins@20211123', retriever: legacySCM(scm))
+    def lib = library(identifier: 'jenkins@1.0.0', retriever: legacySCM(scm))
```
gaiksaya, line 9: Can you please check if `@` is of any relevance for local files/libraries in jenkins? You can find more information here: https://www.jenkins.io/doc/book/pipeline/shared-libraries/#legacy-scm

gaiksaya (changes requested):
I believe this should not be `1.0.0`. More thoughts here https://github.com/opensearch-project/opensearch-build-libraries/issues/7#issuecomment-1259999902

vars/uploadArtifacts.groovy
jordarlu: Thanks for the comment, @gaiksaya .. if you don't mind I put 'jenkins@main' for shared libs under /vars, so, it states the use of main branch clearly. Then, if anyone wants to override it, they can put override in each individual jenkinsfile as your comment at https://github.com/opensearch-project/opensearch-build-libraries/issues/7#issuecomment-1259999902

gaiksaya: Thats for jenkinsFile. However for libraries under `vars` we want it to be the same branch. For example, if the user is using `2.0.0` of this repository the libraries should check out `2.0.0` [here](https://github.com/opensearch-project/opensearch-build-libraries/pull/6/files#diff-4fc12a8f236ed19ed2a263445f244284dc46fdbc62576463d1bd26ba05c9de8fR10) and not main which might point to 3.0 and have a breaking change.

```
lib = library(identifier: 'jenkins@2.0.0', retriever: legacySCM(scm))
```
One way I can think storing that value in `version.properties` and loading it or need to check if above `legacySCM` has better way to checkout current commit than branches.

jordarlu: Crystal clear to me now, makes sense and let me check and update. Thanks again @gaiksaya .

gaiksaya (approved):
Unblocking the ci failures in build repo. @jordarlu Please work on https://github.com/opensearch-project/opensearch-build-libraries/pull/6#discussion_r982887873 in new PR. Thanks!

PR #21, opened 2022-10-25
Create publishToNpm library

gaiksaya:
### Description
Adds publishToNpm library that can be used to publish packages to NPM under `@opensearch-project` name space.
The library takes 2 inputs which is Github repository and the tag

### Issues Resolved
resolves https://github.com/opensearch-project/opensearch-build/issues/2637


vars/standardReleasePipeline.groovy
```diff
@@ -40,7 +40,6 @@ void call(Map args = [:], Closure body) {
             always {
                 script {
                     postCleanup()
-                    sh 'docker image prune -f --all'
```
peterzhuamazon, line 43: Why remove this?

vars/publishToNpm.groovy
```diff
@@ -0,0 +1,22 @@
...
+@param Map args = [:] args A map of the following parameters
+@param args.repository <required> - Repository to be used to publish the artifact to npm
+@param args.tag <required> - Tag reference to be used to publish the artifact
+*/
+void call(Map args = [:]) {
+
+    checkout([$class: 'GitSCM', branches: [[name: "${args.tag}" ]], userRemoteConfigs: [[url: "${args.repository}" ]]])
+    
+    withCredentials([string(credentialsId: 'publish-to-npm-token', variable: 'NPM_TOKEN')]){
+        sh """npm set registry "https://registry.npmjs.org"; npm set //registry.npmjs.org/:_authToken ${NPM_TOKEN}; npm publish --dry-run && npm publish --access public"""
```
peterzhuamazon, line 20: Prefer writing this in multiple lines for easier visibility.

```
sh """
    <code>
"""
```

vars/standardReleasePipeline.groovy
gaiksaya: Caused an error looking for docker command on agent node.

vars/publishToNpm.groovy
gaiksaya: Makes its difficult to test things due to whitespace. Hence used the above approach. Similar to https://github.com/opensearch-project/opensearch-build-libraries/blob/main/vars/copyContainer.groovy#L35

vars/standardReleasePipeline.groovy
peterzhuamazon: Will this affect any other jenkinsfile that calls this?
Ex: We want to do the cleanup for docker build.

vars/publishToNpm.groovy
peterzhuamazon: Huh, I do find it not ideal for reading, what would happen if the code block is too long to put into one line?

vars/standardReleasePipeline.groovy
gaiksaya: No, this file is not used anywhere today.

vars/publishToNpm.groovy
gaiksaya: This was not long enough that's why used this way. If its very long we can use the approach you suggested or try to break it down.

vars/standardReleasePipeline.groovy
rishabh6788: shouldn't agent node supposed to have docker installed? @gaiksaya

peterzhuamazon: @rishabh6788 this file is using docker container running on a jenkins linux agent.
The agent can have docker but the container doesnt have to.
Except docker builder.

rishabh6788: understood, thank you!

vars/publishToNpm.groovy
zelinh: I think from my experience, we used to publish the client to staging NPM before final publish. Would that be performed on the Jenkinsfile level which calls this library?

zelinh, line 20: Maybe we could add one another command `npm logout` at the beginning to make sure we don't accidentally publish to somewhere else or this could be replaced with new login?

gaiksaya: We are using a token and not logging into the account so no need to log out. The scope of scope is within the `withCredentials`.

gaiksaya:
>I think from my experience, we used to publish the client to staging NPM before final publish. Would that be performed on the Jenkinsfile level which calls this library?

@zelinh We won't be dealing with staging anyway. If component team wants to follow that they can do on their own. We will only be directly releasing on prod platform

vars/publishToNpm.groovy
gaiksaya: @zelinh We won't be dealing with staging. If component team wants to follow that they can do on their own. We will only be directly releasing on prod platform

peterzhuamazon (approved):

PR #32, opened 2022-11-10
Add downloading release asset feature 

gaiksaya:
### Description
This PR adds the following:
* Download release asset when `downloadReleaseAsset: true`
* Make the asset name dynamic using `downloadReleaseAssetName`. Defaults to `'artifacts.tar.gz`
* Fix the release trigger where the workflow will be triggered only when a release is `created` and not when its edited or deleted, etc.

### Issues Resolved


vars/standardReleasePipelineWithGenericTrigger.groovy
```diff
@@ -37,21 +38,51 @@ void call(Map args = [:], Closure body) {
...
+                when {
+                    expression {
+                        return args.downloadReleaseAsset
+                    }
+                }
+                steps {
+                    script {
+                        if (args.downloadReleaseAsset && "$assets_url" != '') {
+                            withCredentials([usernamePassword(credentialsId: 'jenkins-github-bot-token', usernameVariable: 'GITHUB_USER', passwordVariable: 'GITHUB_TOKEN')]) {
+                                String assets = sh(
```
peterzhuamazon, line 68: A bit confused what is the need for assets if value can be retrieved from assetUrl?

Thanks

vars/standardReleasePipelineWithGenericTrigger.groovy
```diff
@@ -37,21 +38,51 @@ void call(Map args = [:], Closure body) {
...
+                                )
+                                String assetUrl = null
+                                def parsedJson = readJSON text: assets
+                                parsedJson.each { item ->
+                                    if(item.name == "artifacts.tar.gz") {
+                                        assetUrl = item.url
+                                        }
+                                    }
+                                echo "Downloading artifacts from $assetUrl"
+                                sh "curl -OJ -L -H 'Accept: application/octet-stream' -H 'Authorization: Bearer ${GITHUB_TOKEN}' ${assetUrl} && tar -zxf artifacts.tar.gz"
```
peterzhuamazon, line 80: Instead of doing this you can simply `curl <> | tar -xzv`.

Tar is able to directly extract from stream.

vars/standardReleasePipelineWithGenericTrigger.groovy
gaiksaya: assetUrl is just the api to get all the assets from. Something like: "assets_url": "https://api.github.com/repos/gaiksaya/opensearch-py-ml/releases/82615186/assets"
When you curl this api, this is when you get exact assets like here https://github.com/opensearch-project/opensearch-build-libraries/pull/32/files#diff-cb975467ca0159d0c434e26ed12f26fc453d36cd692d24cf47812eb137f3574bR26

vars/standardReleasePipelineWithGenericTrigger.groovy
rishabh6788: how about `set -ex;curl<>;tar -zxf`? @gaiksaya

gaiksaya: Jenkins / groovy takes care if any shell command fails by default. It won't go on.

peterzhuamazon: @gaiksaya no need, because if curl could not download a file, or download an empty file the return code is always 0.
This is due to cloudfront will return an error file, so curl thinks it downloads the file.

Therefore you need to rely on either the response header code, or tar to detect the errors.
```

$ curl https://ci.opensearch.org/ci/dbc/123
<?xml version="1.0" encoding="UTF-8"?>
<Error><Code>AccessDenied</Code><Message>Access Denied</Message><RequestId>DCFKF4VGS7VNRMMN</RequestId><HostId>Ptk5APk9Cs6M9bbwbGi7UnOvMS0m1hV4tKLcxG2d1N1vkTsW+C1TuLdXfWH+WgG18WFRhXm57fY=</HostId></Error>

$ echo $?
0
```

vars/standardReleasePipelineWithGenericTrigger.groovy
peterzhuamazon: Got it thanks.

vars/standardReleasePipelineWithGenericTrigger.groovy
gaiksaya: So the current set up will handle it too right? Trying to understand what difference `|` will make.

peterzhuamazon: Current setup wont.
You curl a wrong url so CF return an error file, and status 0.
&& will then go to tar and fail due to file incorrect and exit 1.
You leave a wrong file on the server.

My setup is: You curl a wrong url so CF return an error file, and status 0.
Before the file is even fully pulled, tar already try to extract it on the fly, fail due to file incorrect and exit 1.
No leftover on the server and can save space in the extraction process due to no need to download a file beforehand, everything is happening in stream and pipe.

It is much cleaner and runs better when you deal with large files when github actions only have 14G storage.
Tho I wont block you because of this.

gaiksaya: I see. The file won't stay on server as everything is run on docker image. But I get your point. Let me test this in my current set up to see if it works as expected. Will update this thread.
Thanks!

gaiksaya: Hi @peterzhuamazon the mentioned set up does not work. Throws below error:
```
+ tar -zxf artifacts.tar.gz
tar (child): artifacts.tar.gz: Cannot open: No such file or directory
tar (child): Error is not recoverable: exiting now
tar: Child returned status 2
tar: Error is not recoverable: exiting now
```

Will keep what I have as that seems to work as expected. Thanks!

peterzhuamazon: @gaiksaya you misused commands as `-f` is no need anymore due to no files.
Just `tar -xzv` is enough right here.

gaiksaya: Made the changes @peterzhuamazon Thanks for the suggestion. Now the user can use any artifact tarball name. I added one more option. Please review,

peterzhuamazon (approved):

vars/standardReleasePipelineWithGenericTrigger.groovy
peterzhuamazon: Thanks approved.

prudhvigodithi:
Hey @gaiksaya can we have the Defaults to `publish-artifacts.tar.gz`, so that it makes more clear if a user did not pass any `downloadReleaseAssetName`?

gaiksaya:
Sure, I'll create another PR to add this. Wondering if `-` will cause any issues.

PR #56, opened 2022-11-30
Fix verification of signed gems

gaiksaya:
### Description
The `-P HighSecurity` option checks that even dependent gems are signed which might not be necessarily true. The `MediumSecuriy` checks the signatures for signed gems. For unsigned ones, it just ignores.
Hence following this method where we install the gems normally first so that all the dependencies get installed and then uninstall the gem that is to be published and reinstall using `-P HighSecurity`

### Issues Resolved

### DIFF
--- a/tests/jenkins/TestPublishToRubyGems.groovy
+++ b/tests/jenkins/TestPublishToRubyGems.groovy
@@ -12,6 +12,7 @@ package jenkins.tests
 import jenkins.tests.BuildPipelineTest
 import static com.lesfurets.jenkins.unit.MethodCall.callArgsToString
 import static org.hamcrest.CoreMatchers.hasItem
+import static org.hamcrest.CoreMatchers.containsString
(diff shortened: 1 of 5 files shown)


tests/jenkins/TestPublishToRubyGems.groovy
```diff
@@ -12,6 +12,7 @@ package jenkins.tests
 import jenkins.tests.BuildPipelineTest
 import static com.lesfurets.jenkins.unit.MethodCall.callArgsToString
 import static org.hamcrest.CoreMatchers.hasItem
+import static org.hamcrest.CoreMatchers.containsString
```
peterzhuamazon, line 15: Where is this being used in the code?

gaiksaya: Removed! Thanks! Was trying something.

peterzhuamazon (approved):

PR #95, opened 2022-12-13
Remove repo drop logic and add logic to load custom scripts

rishabh6788:
Signed-off-by: Rishabh Singh <sngri@amazon.com>

### Description
Remove repo drop logic as the staging repo is getting auto dropped during release. 

### Issues Resolved


gaiksaya:
[Offline discussion] We also need the publishToMaven library to take care of publishing snapshots.

Approach would be add snapshot: true/false parameter to existing library and call the [publish-snapshot.sh](https://github.com/opensearch-project/opensearch-build-libraries/blob/main/publish/publish-snapshot.sh) script if that's true.

vars/loadCustomScript.groovy
```diff
@@ -0,0 +1,20 @@
...
+/** Library to load custom scripts from resources dir and make them available in the runtime environment to be executed.
+ * See https://www.jenkins.io/doc/book/pipeline/shared-libraries/#loading-resources for more details
+ @param Map args = [:] args A map of the following parameters
+ @param args.scriptPath <required> - Relative Path to the existing script present in resources dir. The path is relative to resources dir. e.g., 'publish/stage-maven-release.sh'
+ @param args.scriptName <optional> - The name of the file that will hold the actual script code, can be same as the existing script name.
+ */
+void call(Map args = [:]) {
+    def scriptContent = libraryResource "${args.scriptPath}"
+    writeFile file: "${args.scriptName}", text: scriptContent
+    sh "chmod a+x ./${args.scriptName}"
```
gaiksaya, line 19: If the scriptName is optional there should be a default right? If not make it required.
Also will the file be written in the current working directory where this library is called?

rishabh6788: I think let us enforce the script name parameter so that user makes sure they are using the right name.
The script is written in the current working directory. @gaiksaya

gaiksaya: Sounds good. Can you change the @param args.scriptName <optional> to @param args.scriptName <required> ?
Thanks!

rishabh6788: Just realized, done! @gaiksaya

gaiksaya (approved):

PR #112, opened 2023-01-19
Limit max workers for windows gradle check.

mch2:
Signed-off-by: Marc Handalian <handalm@amazon.com>

### Description
Windows gradle checks are running into oom issues - see https://build.ci.opensearch.org/job/gradle-check/9515/consoleFull.
I believe this is because the hosts used are limited to 32GB of RAM while spawning threads up to the amount of available cores.  Each worker is allowed max 3GB of ram, so we can easily run oom if too many threads with over 10 workers spawned running expensive tests. This PR limits the max to 8 workers to avoid this scenario.

### Issues Resolved
closes https://github.com/opensearch-project/opensearch-build-libraries/issues/113

### DIFF
--- a/vars/runGradleCheck.groovy
+++ b/vars/runGradleCheck.groovy
@@ -71,7 +71,13 @@ void call(Map args = [:]) {
 
                 echo "Start gradlecheck"
                 GRADLE_CHECK_STATUS=0
-                ./gradlew clean && ./gradlew check -Dtests.coverage=true --no-daemon --no-scan || GRADLE_CHECK_STATUS=1
+                
+                if (uname -s | grep -i NT); then
+                   echo "Start gradlecheck on windows"
+                   ./gradlew clean && ./gradlew check --max-workers 8 -Dtests.coverage=true --no-daemon --no-scan || GRADLE_CHECK_STATUS=1
+                else
+                   ./gradlew clean && ./gradlew check -Dtests.coverage=true --no-daemon --no-scan || GRADLE_CHECK_STATUS=1
+                fi
 
                 if [ "\$GRADLE_CHECK_STATUS" != 0 ]; then
                     echo Gradle Check Failed!
--- a/tests/jenkins/jobs/RunGradleCheck_Jenkinsfile.txt
+++ b/tests/jenkins/jobs/RunGradleCheck_Jenkinsfile.txt
@@ -57,7 +57,13 @@
 
                 echo "Start gradlecheck"
                 GRADLE_CHECK_STATUS=0
-                ./gradlew clean && ./gradlew check -Dtests.coverage=true --no-daemon --no-scan || GRADLE_CHECK_STATUS=1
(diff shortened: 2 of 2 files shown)


vars/runGradleCheck.groovy
```diff
@@ -71,7 +71,13 @@ void call(Map args = [:]) {
 
                 echo "Start gradlecheck"
                 GRADLE_CHECK_STATUS=0
-                ./gradlew clean && ./gradlew check -Dtests.coverage=true --no-daemon --no-scan || GRADLE_CHECK_STATUS=1
+                
+                if (uname -s | grep -i NT); then
+                   echo "Start gradlecheck on windows"
+                   ./gradlew clean && ./gradlew check --max-workers 8 -Dtests.coverage=true --no-daemon --no-scan || GRADLE_CHECK_STATUS=1
```
gaiksaya, line 77: Can the max-workers change in future? Asking if it can be something that should be passed as an argument?

gaiksaya (commented):
Looks like tests are failing! Not sure if you ran `./gradlew test -info -Ppipeline.stack.write=true` so that txt file gets update accordingly.

vars/runGradleCheck.groovy
mch2: yes that would be ideal to pass the entire command from https://github.com/opensearch-project/OpenSearch/blob/main/.github/workflows/gradle-check.yml. This change is against windows CI only that runs nightly & that config is not in the other repo, so we'll also want to move that before being able to pass as a param bc we don't want to impact non windows checks.

gaiksaya: Maybe time to redo gradle check workflow as well? :) (cc: @bbarani)

Looks like it is distributed in around 3 repos today which is 
Build repo: 
https://github.com/opensearch-project/opensearch-build/blob/main/scripts/gradle/gradle-check.sh
https://github.com/opensearch-project/opensearch-build/blob/main/jenkins/gradle/gradle-check.jenkinsfile

This repo and OpenSearch repo.

Adding @peterzhuamazon to this conversation as he is the original contributor.

peterzhuamazon: Thanks @gaiksaya I will work with @mch2 on this.

peterzhuamazon:
Another way is to add another profile just for gradle check AMI in opensearch-ci.

mch2:
> Another way is to add another profile just for gradle check AMI in opensearch-ci.

Just cut this https://github.com/opensearch-project/opensearch-build-libraries/issues/114

peterzhuamazon:
Pausing this a bit as we are testing the bigger winrm memory setups here:
* https://github.com/opensearch-project/opensearch-ci/pull/242

mch2:
Closing this as memory issues were resolved by https://github.com/opensearch-project/opensearch-ci/pull/242

PR #117, opened 2023-01-28
Add publishToNuget lib and fix workspace overwrite issue for signing

gaiksaya:
Signed-off-by: Sayali Gaikawad <gaiksaya@amazon.com>

### Description
- Adds a new publishToNuget library that can be used to publish dotnet components Nuget Gallery
- Also fixes and resolves https://github.com/opensearch-project/opensearch-build/issues/1436

### Issues Resolved
resolves  https://github.com/opensearch-project/opensearch-build/issues/1436
resolves https://github.com/opensearch-project/opensearch-build/issues/2965


tests/jenkins/jobs/PublishToNuget_Jenkinsfile
```diff
@@ -0,0 +1,26 @@
...
+
+pipeline {
+    agent none
+    stages {
+        stage('publishToNuget') {
+            steps {
+                script {
+                    publishToNuget(
+                        repository: 'https://github.com/opensearch-project/opensearch-net',
+                        tag: '1.2.0',
```
prudhvigodithi, line 18: Hey @gaiksaya this tag would be inferred from the GH web-hook data right and parsed to the Jenkinsfile?

gaiksaya: Yes! https://github.com/opensearch-project/opensearch-build-libraries/blob/main/vars/standardReleasePipelineWithGenericTrigger.groovy#L57

prudhvigodithi:
@gaiksaya LGTM, only code we have to look for is `for package in `find src -name OpenSearch*.nupkg``, there is no chance the package that we publish will be without `OpenSearch` prefix as we allow to publish only with `OpenSearch` prefix to the dotnet account, but I'm just thinking if any dependent packages exists without `OpenSearch` prefix, if so we should work with dotnet repo team.

prudhvigodithi (approved):

gaiksaya:
> @gaiksaya LGTM, only code we have to look for is `for package in `find src -name OpenSearch*.nupkg``, there is no chance the package that we publish will be without `OpenSearch` prefix as we allow to publish only with `OpenSearch` prefix to the dotnet account, but I'm just thinking if any dependent packages exists without `OpenSearch` prefix, if so we should work with dotnet repo team.

Even if there are I believe we should not be signing them as our own. @Xtansia can confirm though is there is any possibility in future.

prudhvigodithi:
> Even if there are I believe we should not be signing them as our own. @Xtansia can confirm though is there is any possibility in future.

Since those dependent packages will be inside the` .nukpg` file, its better we sign them as well as they are indirectly used by the main package

Xtansia:
> > Even if there are I believe we should not be signing them as our own. @Xtansia can confirm though is there is any possibility in future.
> 
> Since those dependent package `dll's` will be inside the` .nukpg` file, its better we sign them as well as they are indirectly used by the main package

The NuGet package only contains the DLL for the specific library, it doesn't contain dependencies DLLs, it's purely managed via metadata.
For example, the `OpenSearch.Client` library depends on the `OpenSearch.Net` library (which has it's own dependencies), but the `OpenSearch.Client.x.y.z.nupkg` only contains the `OpenSearch.Client.dll` (for each framework), and the `OpenSearch.Net.x.y.z.nupkg` only contains the `OpenSearch.Net.dll` (for each framework).

This can be seen here: https://nuget.info/packages/OpenSearch.Client/1.2.0

Xtansia:
And yes, all publicly consumable packages from `opensearch-net` do and will have the `OpenSearch.` namespace.

gaiksaya:
> And yes, all publicly consumable packages from `opensearch-net` do and will have the `OpenSearch.` namespace.

Thanks! That answers our question.

PR #119, opened 2023-02-01
Support switching ruby gems version and fix regex

gaiksaya:
Signed-off-by: Sayali Gaikawad <gaiksaya@amazon.com>

### Description
The recent run for rubyGems failed for logstash-output-opensearch because instead of ruby it is build on jruby. https://build.ci.opensearch.org/job/logstash-output-opensearch-release/3/console
In order to fix that, this change adds the enhancement to install any ruby or jruby version as a parameter and installs it on the go.
It is required only for gem signature validation and not for building. 
This change also fixes the regex which is used to retrieve the gem name. 

### Issues Resolved
related: https://github.com/opensearch-project/opensearch-build/issues/3139

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -120,7 +120,7 @@ jacocoTestReport {
     }
 }
 
-String version = '1.5.4'
+String version = '1.5.5'
 
 task updateVersion {
     doLast {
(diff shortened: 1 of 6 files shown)


sshivanii (approved):
Thanks!

peterzhuamazon (approved):

peterzhuamazon (approved):

PR #168, opened 2023-03-15
Fix publishing to Nuget syntax and refactor find

gaiksaya:
### Description
The recent release for dot net client failed because of bash syntax issue. This change is trying to use bash as much less as possible. Also refactored the code a bit as suggested by dotnet team so that only the necessary DLLs are signed, as the DLLs get duped between bin & obj folders (and between projects) during build we only want to sign the one from bin/Release folder. 

### Issues Resolved

### DIFF
--- a/vars/publishToNuget.groovy
+++ b/vars/publishToNuget.groovy
@@ -19,9 +19,10 @@
 void call(Map args = [:]) {
     lib = library(identifier: 'jenkins@main', retriever: legacySCM(scm))
     checkout([$class: 'GitSCM', branches: [[name: "${args.tag}" ]], userRemoteConfigs: [[url: "${args.repository}" ]]])
+
     sh """
     dotnet build ${WORKSPACE}/${args.solutionFilePath} --configuration Release
-    find src -name OpenSearch*.dll>${WORKSPACE}/dlls.txt
+    find src/OpenSearch.*/bin/Release/*/*.dll -type f -regextype posix-extended -regex "src/([^/]+)/bin/Release/[^/]+/\\1\\.dll">${WORKSPACE}/dlls.txt
     """
     dlls = readFile(file: "${WORKSPACE}/dlls.txt").readLines()
     dlls.each { item ->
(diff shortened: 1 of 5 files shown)


vars/publishToNuget.groovy
```diff
@@ -31,13 +32,16 @@ void call(Map args = [:]) {
             overwrite: true
             )
         }
+
+    sh """
+        dotnet pack ${WORKSPACE}/${args.solutionFilePath} --configuration Release --no-build
+        find src -name OpenSearch*.nupkg > ${WORKSPACE}/nupkg.txt
```
peterzhuamazon, line 38: Just wondering do we have to use a file instead of saving in var?

vars/publishToNuget.groovy
```diff
@@ -19,9 +19,10 @@
 void call(Map args = [:]) {
     lib = library(identifier: 'jenkins@main', retriever: legacySCM(scm))
     checkout([$class: 'GitSCM', branches: [[name: "${args.tag}" ]], userRemoteConfigs: [[url: "${args.repository}" ]]])
+
     sh """
     dotnet build ${WORKSPACE}/${args.solutionFilePath} --configuration Release
-    find src -name OpenSearch*.dll>${WORKSPACE}/dlls.txt
+    find src/OpenSearch.*/bin/Release/*/*.dll -type f -regextype posix-extended -regex "src/([^/]+)/bin/Release/[^/]+/\\1\\.dll">${WORKSPACE}/dlls.txt
```
peterzhuamazon, line 25: Can you explain what this regex is trying to do?
Thanks.

peterzhuamazon (approved):
Nitpik mostly.
Thanks.

vars/publishToNuget.groovy
gaiksaya: Makes it easier to read a file in groovy that var.

vars/publishToNuget.groovy
gaiksaya: Get all OpenSearch dll files that needs to be signed. Looks like dlls are duplicated across the folders while building so we just want to sign want will be released. I got this from @Xtansia reason is mentioned in PR description.

Xtansia: For context the "true" DLLs that needs signing and that will be packaged, for each project is:
`src/{project_name}/bin/Release/{target_framework}/{project_name}.dll`
But all the dependency DLLs also get copied into `bin/Release/{target_framework}` and there's also copies in an `obj` so can't just blindly take all DLLs. So this regex captured the `{project_name}` and uses a back reference `\1` to make sure it finds the DLL with the same name as the project.
You can see the difference in output between the previous version and this one in this screenshot:
Screenshot 2023-03-15 at 12 53 14 PM

vars/publishToNuget.groovy
peterzhuamazon: ok sure.

vars/publishToNuget.groovy
peterzhuamazon: Got it, thanks both for the great explanation.

PR #177, opened 2023-04-05
Support tarball publishing for npm

gaiksaya:
### Description
Adds support to publish artifact as part of the publishing. More details: https://docs.npmjs.com/cli/v9/commands/npm-publish?v=true#description
This new change is breaking hence bumping the version to 3.0.0.
This change introduces a `publicationtype` as either `github` in which case the github url and tag is retrieved from the environment. In this case if is not declared the checkout fails which we want. 
Another `publicationtype` is `artifact` that needs `artifactPath`. If either of the publicationType is not passed it hard fails as we do not want the default workspace (containing package.json) to be published. 

### Issues Resolved
resolves https://github.com/opensearch-project/opensearch-build/issues/3312


vars/publishToNpm.groovy
```diff
@@ -7,16 +7,30 @@
...
 void call(Map args = [:]) {
+    artifactPath = args.artifactPath ?: ''
 
-    checkout([$class: 'GitSCM', branches: [[name: "${args.tag}" ]], userRemoteConfigs: [[url: "${args.repository}" ]]])
-    
-    withCredentials([string(credentialsId: 'jenkins-opensearch-publish-to-npm-token', variable: 'NPM_TOKEN')]){
-        sh """npm set registry "https://registry.npmjs.org"; npm set //registry.npmjs.org/:_authToken ${NPM_TOKEN}; npm publish --dry-run && npm publish --access public"""
+    if (args.artifactPath && args.repository && args.tag) {
+        currentBuild.result = 'ABORTED'
+        error('Artifact Path and GitHub repository details are mutually exclusive. Please use either repository and tag OR artifactPath as arguments')
```
peterzhuamazon, line 22: Is it possible to use a param to define whether it is artifact path or github repository?
Like github=true.
Instead of doing this cumbersome way to switch?

vars/publishToNpm.groovy
```diff
@@ -7,16 +7,30 @@
...
 
-/** Library to publish artifacts to NPM registry under @opensearch-project namespace
+/** Library to publish artifacts to NPM registry under @opensearch-project namespace.
 @param Map args = [:] args A map of the following parameters
-@param args.repository <required> - Repository to be used to publish the artifact to npm
-@param args.tag <required> - Tag reference to be used to publish the artifact
+@param args.repository <required> - Repository to be used to publish the artifact to npm.
+@param args.tag <required> - Tag reference to be used to publish the artifact.
+OR
+@param args.artifactPath <required> - Artifact path to publish to NPM. Defaults to package.json See supported artifacts https://docs.npmjs.com/cli/v9/commands/npm-publish?v=true#description for more details.
```
peterzhuamazon, line 15: See next comment on my take on this.

vars/publishToNpm.groovy
```diff
@@ -7,16 +7,30 @@
...
-        sh """npm set registry "https://registry.npmjs.org"; npm set //registry.npmjs.org/:_authToken ${NPM_TOKEN}; npm publish --dry-run && npm publish --access public"""
+    if (args.artifactPath && args.repository && args.tag) {
+        currentBuild.result = 'ABORTED'
+        error('Artifact Path and GitHub repository details are mutually exclusive. Please use either repository and tag OR artifactPath as arguments')
     }
-}
\ No newline at end of file
+    if (!args.artifactPath) {
+        checkout([$class: 'GitSCM', branches: [[name: "${args.tag}" ]], userRemoteConfigs: [[url: "${args.repository}" ]]])
+    }
```
peterzhuamazon, line 26: if github=true; checkout, else, no.

vars/publishToNpm.groovy
```diff
@@ -7,16 +7,30 @@
...
+
+    withCredentials([string(credentialsId: 'jenkins-opensearch-publish-to-npm-token', variable: 'NPM_TOKEN')]) {
+        sh """
+            npm set registry "https://registry.npmjs.org"
+            npm set //registry.npmjs.org/:_authToken ${NPM_TOKEN}
```
peterzhuamazon, line 31: How to unset the token after the publish?
Is there any file deletion/cleaning or logout method?

vars/publishToNpm.groovy
```diff
@@ -7,16 +7,30 @@
...
+            npm publish ${artifactPath} --dry-run
```
peterzhuamazon, line 32: Is it possible to use dryrun to test whether it is successful?
Then could you add things like `set -e` to exit if dryrun returns non-zero status?
Thanks.

vars/publishToNpm.groovy
```diff
@@ -7,16 +7,30 @@
...
+        error('Artifact Path and GitHub repository details are mutually exclusive. Please use either repository and tag OR artifactPath as arguments')
     }
-}
\ No newline at end of file
+    if (!args.artifactPath) {
+        checkout([$class: 'GitSCM', branches: [[name: "${args.tag}" ]], userRemoteConfigs: [[url: "${args.repository}" ]]])
+    }
+
+    withCredentials([string(credentialsId: 'jenkins-opensearch-publish-to-npm-token', variable: 'NPM_TOKEN')]) {
+        sh """
```
peterzhuamazon, line 29: Might want to use nvm to switch node/npm version just in case a bump to the image, then cause issues during the publish.

peterzhuamazon (changes requested):
Hi @gaiksaya ,

Have some comments.

Thanks.

vars/publishToNpm.groovy
gaiksaya: Node is not in picture at all since we are not building anything on jenkins so we are good for now.

vars/publishToNpm.groovy
gaiksaya: I can add `&&` if that's okay. That is what was before. So next command will be executed only when dry run is successful

vars/publishToNpm.groovy
peterzhuamazon: Sure

vars/publishToNpm.groovy
peterzhuamazon: Sure.

vars/publishToNpm.groovy
peterzhuamazon: I am actually not sure if nvm use change to a different node, does npm change as well or not.

gaiksaya: I dont think so. since publishing is not as such building anything any node should be fine. Has not caused any issues yet but will keep watch.

vars/publishToNpm.groovy
gaiksaya: Addressed in recent commit. Please check

peterzhuamazon: Thanks.

vars/publishToNpm.groovy
```diff
@@ -7,16 +7,34 @@
...
-        sh """npm set registry "https://registry.npmjs.org"; npm set //registry.npmjs.org/:_authToken ${NPM_TOKEN}; npm publish --dry-run && npm publish --access public"""
+    if (!allowedPublicationType.contains(args.publicationType)) {
+        currentBuild.result = 'ABORTED'
+        error('Invalid publicationType. publicationType can either be github or artifact')
     }
+    if (args.publicationType == 'artifact' && !args.artifactPath) {
+        currentBuild.result = 'ABORTED'
+        error('publicationType: artifact needs an artifactPath. Please provide artifactPath argument. See supported artifacts https://docs.npmjs.com/cli/v9/commands/npm-publish?v=true#description for more details')
+    }
+    if (args.publicationType == 'github') {
```
peterzhuamazon, line 27: You probably also want to check if type is github, then artifactpath needs to be empty or the npm publish will go back to publishing with tar.

gaiksaya: Thanks for bringing this up. Added the check. Created a new method just for parameter check to make it cleaner.

peterzhuamazon (approved):

PR #187, opened 2023-04-11
Add --all-tags copy of container image between different container registries

peterzhuamazon:
### Description
Add --all-tags copy of container image between different container registries.
Either between docker hub / ecr or other registries.

### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/3386


gaiksaya (changes requested):
I am still not clear (after reading related GH issue) what we are trying to achieve here? When you say recursive copy, it is just for one image or all images? For example we build os-ubuntu-1 and it pushed to dockerhub, enabling recursive copy will push to ECR too simultaneously?

vars/copyContainer.groovy
```diff
@@ -10,34 +10,67 @@
...
     if(args.destinationRegistry == 'public.ecr.aws/opensearchstaging') {
-            def ecrLogin = sh(returnStdout: true, script: "set +x && aws ecr-public get-login-password --region us-east-1 | docker login --username AWS --password-stdin ${args.destinationRegistry}").trim()
-            sh """gcrane cp ${args.sourceRegistry}/${args.sourceImage} ${args.destinationRegistry}/${args.destinationImage}; docker logout ${args.destinationRegistry}"""
+        def ecrLogin = sh(returnStdout: true, script: "set +x && aws ecr-public get-login-password --region us-east-1 | docker login --username AWS --password-stdin ${args.destinationRegistry}").trim()
+        gcraneCopy()
     }
+
+}
+
+def gcraneCopy() {
```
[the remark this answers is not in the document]
peterzhuamazon: There are at least 4 vars need to be passed that is why I use global.

tests/jenkins/jobs/DockerCopyRecursive_Jenkinsfile
```diff
@@ -0,0 +1,42 @@
...
+                                        error('Make sure all the parameters are passed in.')
+                            }
+                        }
+                    }
+                }
+                stage('Copy Image to ECR/DockerHub Recursively') {
+                    steps {
+                        script {
+                            copyContainer(
+                                recursiveCopy: "true",
```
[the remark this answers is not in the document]
gaiksaya: Most libraries are using boolean so please use boolean. Also makes it easy for other groovy methods and loops.

vars/copyContainer.groovy
```diff
@@ -10,34 +10,67 @@
...
+
+        if [ ${recursive_copy} = 'true' ]; then
+            echo "Recursive copy from ${source_registry}/${source_image_no_tag} to ${destination_registry}/${destination_image_no_tag}"
+            for source_entry in `gcrane ls ${source_registry}/${source_image_no_tag}`; do
+                image_tag=`echo \$source_entry | cut -d/ -f3 | cut -d: -f2`
+                destination_entry="${destination_registry}/${destination_image_no_tag}:\$image_tag"
+                gcrane cp \$source_entry \$destination_entry
+            done
+        else
+            echo "Normal copy from ${source_registry}/${source_image} to ${destination_registry}/${destination_image}"
```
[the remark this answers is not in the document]
peterzhuamazon: Added now.

vars/copyContainer.groovy
```diff
@@ -10,34 +10,67 @@
...
+        gcraneCopy()
     }
+
+}
+
+def gcraneCopy() {
+    sh """
+        set +x
+
+        if [ ${recursive_copy} = 'true' ]; then
```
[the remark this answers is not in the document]
peterzhuamazon: This is fine as not only `gcrane cp` but also `gcrane ls` and tags processing can be added in shell.
We can improve this later with a additional PR.

Thanks.

vars/copyContainer.groovy
```diff
@@ -10,34 +10,67 @@
...
  * @param args A map of the following parameters
+ * @param args.recursiveCopy Copy all the tags of a sourceImage to destinationImage if 'true' and <IMAGE_TAG> is ignored in sourceImage/destinationImage, default to 'false'
  * @param args.sourceImage The Source Image name and tag <IMAGE_NAME>:<IMAGE_TAG> Eg: opensearch:1.3.2
  * @param args.sourceRegistry The source docker registry, currently supports 'DockerHub' or 'ECR'
  * @param args.destinationImage The Destination Image name and tag <IMAGE_NAME>:<IMAGE_TAG> Eg: opensearch:1.3.2
  * @param args.destinationRegistry The destination docker registry, currently supports 'DockerHub' or 'ECR'
  */
 void call(Map args = [:]) {
 
+    recursive_copy = args.recursiveCopy
```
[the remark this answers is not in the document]
peterzhuamazon: Added now.

tests/jenkins/jobs/DockerCopyRecursive_Jenkinsfile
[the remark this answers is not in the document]
peterzhuamazon: Used boolean now.

vars/copyContainer.groovy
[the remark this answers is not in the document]
rishabh6788: Have limited understanding of how groovy deals with variable scope so my only concern is that is the variables declared in `call()` scope accessible in gcraneCopy() scope? @peterzhuamazon

vars/copyContainer.groovy
```diff
@@ -10,34 +10,67 @@
...
+        gcraneCopy()
     }
+
+}
+
+def gcraneCopy() {
+    sh """
+        set +x
+
+        if [ ${recursive_copy} = true ]; then
```
rishabh6788, line 60: Have limited understanding of how groovy deals with variable scope so my only concern is that is the variables declared in call() scope accessible in gcraneCopy() scope? @peterzhuamazon. 
If you have already tested this then I am fine.

vars/copyContainer.groovy
```diff
@@ -10,34 +10,67 @@
...
             {
                 withAWS(role: "${ARTIFACT_PROMOTION_ROLE_NAME}", roleAccount: "${AWS_ACCOUNT_ARTIFACT}", duration: 900, roleSessionName: 'jenkins-session') {
                     def ecrLogin = sh(returnStdout: true, script: "set +x && aws ecr-public get-login-password --region us-east-1 | docker login --username AWS --password-stdin ${args.destinationRegistry}").trim()
-                    sh """gcrane cp ${args.sourceRegistry}/${args.sourceImage} ${args.destinationRegistry}/${args.destinationImage}; docker logout ${args.destinationRegistry}"""
+                    gcraneCopy()
                 }
             }
     }
+
     if(args.destinationRegistry == 'public.ecr.aws/opensearchstaging') {
```
rishabh6788, line 49: Why is this different from `public.ecr.aws/opensearchproject` in terms of logging in?

vars/copyContainer.groovy
```diff
@@ -10,34 +10,67 @@
  * Copies a container from one docker registry to another
  *
  * @param args A map of the following parameters
+ * @param args.recursiveCopy Copy all the tags of a sourceImage to destinationImage if 'true' and <IMAGE_TAG> is ignored in sourceImage/destinationImage, default to 'false'
  * @param args.sourceImage The Source Image name and tag <IMAGE_NAME>:<IMAGE_TAG> Eg: opensearch:1.3.2
```
rishabh6788, line 14: could you add which parameters are required and which are optional?

vars/copyContainer.groovy
peterzhuamazon: It is global variable so `gcraneCopy()` can easily access the variable defined in `call()`.
We could probably make a smaller scope access with parameter passing, but that can be a later improvement.
Thanks.

peterzhuamazon: And yes it is tested in staging.

vars/copyContainer.groovy
peterzhuamazon: That is because the staging ecr and prod ecr are in 2 different accounts.
So different credentials.
Thanks.

vars/copyContainer.groovy
peterzhuamazon: Added now.

vars/copyContainer.groovy
```diff
@@ -10,34 +10,67 @@
...
     }
+
+}
+
+def gcraneCopy() {
+    sh """
+        set +x
+
+        if [ ${recursive_copy} = true ]; then
+            echo "Copying all image tags recursively from ${source_registry}/${source_image_no_tag} to ${destination_registry}/${destination_image_no_tag}"
```
rishabh6788, line 61: If I understand correctly at any given moment the recursive copy will only happen between values provided in `source_registry` and `destination_registry`. Meaning if the values are `opensearchproject` and `opensearchstaging` then all the tags for a given image will get copied from `opensearchproject` to `opensearchstaging` but no copy will happen to ecr repo. Right?

rishabh6788: I think my point is very much similar to what @gaiksaya is asking.

rishabh6788 (approved):
As understood, the recursive copy will only happen for tags between provided registries. 
LGTM.

vars/copyContainer.groovy
peterzhuamazon: If you specify source_reg as ecrstaging, and dest_reg as ecrprod, yes it will copy all the tags of a given image from ecrstaging to ecrprod, vise versa. If you specify another registry like gcp container registry it is the same thing, it will copy from one registry to another registry, copy all the tags of a image between given registries of the image.

Also, yeah it is isolated only between the given registries, recursively copy all the tags of a given image, not iterate among all the registries at all.

peterzhuamazon:
Apparently someone has pushed a change last week in the latest version `0.15.x` to allow all tags push during gcrane copy.
* https://github.com/google/go-containerregistry/pull/1682

Which solves: 
* https://github.com/google/go-containerregistry/issues/1355

peterzhuamazon:
Will consider making that change to avoid issues like manually iterate through all the tags by gathering info from `gcrane ls`.

peterzhuamazon:
New commands:

```
./crane <reg>/<repo>/<image> <reg>/<repo>/<image> --all-tags
```

peterzhuamazon:
Pre-requisite: 
* https://github.com/opensearch-project/opensearch-build/pull/3490

tests/jenkins/TestCopyContainer.groovy
```diff
@@ -38,28 +39,34 @@ class TestCopyContainer extends BuildPipelineTest {
...
-         def shellCommands = getCommandExecutions('sh', 'gcrane').findAll {
-            shCommand -> shCommand.contains('gcrane')
-         }
 
-         assertThat(shellCommands.size(), equalTo(1))
-         assertThat(shellCommands, hasItem("gcrane cp opensearchstaging/alpine:3.15.4 opensearchproject/alpine:3.15.4; docker logout".toString()))
+        String crane_str = 'set -x && crane cp opensearchstaging/alpine:3.15.4 opensearchproject/alpine:3.15.4'
+        assertThat(getShellCommands('sh', 'crane'), hasItem(crane_str))
+
+        String docker_str = 'set +x && docker logout && docker logout opensearchproject'
```
gaiksaya, line 46: Didn't get why we are logging out both?

peterzhuamazon:
Hi @gaiksaya @rishabh6788 please review again as the current implementation is much cleaner.

Also:
1. Switch from `gcrane` to `crane` as `--all-tags` functionality is specifically added on `crane`.
2. `gcrane` is just a gcr specific `crane` variant that has better `ls` function, `crane` can do same thing and it is the source of truth. Similarly `krane` is more for the kube variant. See https://github.com/google/go-containerregistry#tools.
3. Make more logic in groovy to replace full-blown shell scripts.

Thanks.

vars/copyContainer.groovy
gaiksaya: This is really risky and with very high blast radius. I don't want anyone mistakenly copying all tags from opensearchstaging to opensearchProject dockerhub. 
One mis-click and all images will be overwritten. Why would we want to copy all tags from one source to destination? One copy at a time is good. I believe we should not add recursive copy instead call this method recursively in jenkins file if required. 

cc: @bbarani @prudhvigodithi Needs your input here.

gaiksaya (changes requested):

vars/copyContainer.groovy
peterzhuamazon: If needed I can move this to its own workflow.
And I think it is very beneficial.

We can have a workflow running on cron to sync images between hub and ecr such like ci-runner images, and we can have a quick sync without human intervention. 

This can also be a active backup function so we dont have to manually do multiple copies just to create one image and forget to backup in ECR or other registry.

We can use the workflow jenkinsfile to lock down source and destination, and this can happen between only staging and staging.

All of our ci images are versioned, means the risk is not as high. All in all, this is to provide a mean for us to easily manage our sync images and backup images.

Have a workflow slowly schedule one to one copy can be very expansive and time consuming, especially if it is for backup images. And we already have a lot of tags right now. It is not going to scale on the long run.

Thanks.

tests/jenkins/TestCopyContainer.groovy
peterzhuamazon: Because this `craneCopy` function is used for not only docker hub but also ecr, logging out both is not going to send error and we make sure we are logging out of either.

peterzhuamazon:
Replied your comment @gaiksaya see my responses:
https://github.com/opensearch-project/opensearch-build-libraries/pull/187#discussion_r1189230870

cc: @bbarani @rishabh6788 @prudhvigodithi .

vars/copyContainer.groovy
gaiksaya: In that case I believe we should only allow recursive for CI-runner images where source and destination can either be `opensearchstaging` and/or `public.ecr.aws/opensearchstaging` and repo name being ci-runner.
The same environment also contains credentials for prod and other opensearch images which we do not want to sync for all tags anytime between staging and Prod for any version. 

Also can you add tests for default and non-default behavior?
Thanks!

vars/copyContainer.groovy
```diff
@@ -9,35 +9,62 @@
...
+        sh("set +x && aws ecr-public get-login-password --region us-east-1 | docker login --username AWS --password-stdin ${args.destinationRegistry}")
+        craneCopy()
+    }
+
+}
+
+def craneCopy() {
+
+    if (all_tags == true) {
+        echo "Copying all image tags from ${source_registry}/${source_image_no_tag} to ${destination_registry}/${destination_image_no_tag}"
```
gaiksaya, line 59:
```suggestion
        echo "Copying all the tags of image ${source_registry}/${source_image_no_tag} to ${destination_registry}/${destination_image_no_tag}"
```

gaiksaya (changes requested):

vars/copyContainer.groovy
peterzhuamazon: We already have tests for default vs non-default behavior, will add restrictions of destination and repo name.

peterzhuamazon: Added now.

rishabh6788:
LGTM

rishabh6788 (approved):

vars/copyContainer.groovy
```diff
@@ -10,34 +10,66 @@
...
+
+def craneCopy() {
+
+    if (all_tags == true) {
+        if (destination_registry.contains('opensearchstaging') && destination_image_no_tag.equals('ci-runner')) {
+            echo "Copying all the tags of image ${source_registry}/${source_image_no_tag} to ${destination_registry}/${destination_image_no_tag}"
+            sh("set -x && crane cp ${source_registry}/${source_image_no_tag} ${destination_registry}/${destination_image_no_tag} --all-tags")
+        }
+        else {
+            error("'destination_registry' must be opensearchstaging resgistry and 'destination_image' must be ci-runner")
```
gaiksaya, line 64:
```suggestion
            error("'destination_registry' must be opensearchstaging registry and 'destination_image' must be ci-runner")
```

vars/copyContainer.groovy
```diff
@@ -10,34 +10,66 @@
...
+        sh("set +x && aws ecr-public get-login-password --region us-east-1 | docker login --username AWS --password-stdin ${args.destinationRegistry}")
+        craneCopy()
+    }
+
+}
+
+def craneCopy() {
+
+    if (all_tags == true) {
+        if (destination_registry.contains('opensearchstaging') && destination_image_no_tag.equals('ci-runner')) {
```
gaiksaya, line 59: Confused with image name and repository name:
https://hub.docker.com/r/opensearchstaging/ci-runner/tags
https://gallery.ecr.aws/opensearchstaging/ci-runner
When you say registry it also includes repo name right?

peterzhuamazon: yes, that is not ideal but it is by design following the initial design of the docker copy workflow.

peterzhuamazon: docker.io/opensearchstaging vs. public.ecr.aws/opensearchstaging, technically this is reg+repo name overall.

gaiksaya: Can we atleast make the variables name right then?
`destination_image_no_tag` -> `repository_name`?

peterzhuamazon: That one is correct var name.
destination image is: `ci-runner:centos-cli`
destination_image_no_tag is: `ci-runner`

You are mistaken the repo name as image name.

peterzhuamazon: public.ecr.aws/opensearchstaging/ci-runner:centos-cli
|---registry---|----repo name----|--image--|---tag---|

gaiksaya (approved):

PR #236, opened 2023-06-28
Fix signing

gaiksaya:
### Description
Windows and macos signing do not need gpg key as they are signed using internal keys. Hence removing the unwanted import from the code.
Also they do not need the signature type as that is not an option. Hence, fixed `publishToArtifactsProdBucket` library with same. 

`signArtifacts` needs more fixing 

### Issues Resolved


vars/publishToArtifactsProdBucket.groovy
```diff
@@ -19,18 +19,26 @@
...
     println('Signing the artifacts')
-    signArtifacts(
+    if (args.signingPlatform == 'windows' || args.signingPlatform == 'mac') {
+        signArtifacts(
+            artifactPath: args.source,
+            platform: args.signingPlatform,
+            overwrite: args.sigOverwrite ?: true
+            )
+    } else {
+        signArtifacts(
```
peterzhuamazon, line 29: This change will just sign `windows` zip with windows signing, which should not be and should just use the normal detached signature method.

Thoughts?

peterzhuamazon (changes requested):

vars/publishToArtifactsProdBucket.groovy
gaiksaya: That is what we want. Windows does not do detached signature. It signs actual binary. Only pgp gives detached signature

peterzhuamazon: I am talking about the windows zip we release alongside linux tar.
https://artifacts.opensearch.org/releases/bundle/opensearch/2.8.0/opensearch-2.8.0-windows-x64.zip
https://artifacts.opensearch.org/releases/bundle/opensearch/2.8.0/opensearch-2.8.0-windows-x64.zip.sig

gaiksaya: That uses PGP signing not windows signing.

peterzhuamazon: Yes that is my point your change will just make the zip not able to be signed with detached sig, aka pgp.

peterzhuamazon: Oh this is not related to artifact promotion right? Only client release?

gaiksaya: How? The library you are looking at is `vars/publishToArtifactsProdBucket.groovy` I did not change anything in `signArtiacts` library.
Also for above artifacts you mentioned. args.signingPlatform == linux

gaiksaya: Yes you are right! We can reuse this library but for now it is not affecting any promoteArifacts workflow

peterzhuamazon: Got it, thanks.

peterzhuamazon (approved):

PR #254, opened 2023-07-26
Make docker command detection consistently across OSes

peterzhuamazon:
### Description
Make docker command detection consistently across OSes

### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/3816

### DIFF
--- a/vars/runGradleCheck.groovy
+++ b/vars/runGradleCheck.groovy
@@ -51,7 +51,7 @@ void call(Map args = [:]) {
                 ./gradlew --stop
                 rm -rf ~/.gradle
 
-                if ! (uname -s | grep -i NT); then
+                if command -v docker > /dev/null; then
                     echo "Check existing dockercontainer"
                     docker ps -a
                     docker stop `docker ps -qa` > /dev/null 2>&1 || echo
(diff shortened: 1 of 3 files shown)


vars/runGradleCheck.groovy
```diff
@@ -51,7 +51,7 @@ void call(Map args = [:]) {
                 ./gradlew --stop
                 rm -rf ~/.gradle
 
-                if ! (uname -s | grep -i NT); then
+                if command -v docker > /dev/null; then
                     echo "Check existing dockercontainer"
                     docker ps -a
```
rishabh6788, line 56: `ps -a` would also list the exited containers which are already stopped. 
If we just want to stop the running containers then we should remove the `-a` flag. 
@peterzhuamazon

vars/runGradleCheck.groovy
```diff
@@ -51,7 +51,7 @@ void call(Map args = [:]) {
...
                     docker stop `docker ps -qa` > /dev/null 2>&1 || echo
```
rishabh6788, line 57: should be use `force` to make sure all the stale containers are removed?
@peterzhuamazon

peterzhuamazon: I will add force but force is more or less trying to rm containers that is still not stopped, considering I already try once to stop the container, this should not be needed but anyway nice to have.

Thanks.

rishabh6788 (approved):

PR #268, opened 2023-08-01
Docker Patch Library to support Docker re-release Automation 

Divyaasm:
### Description
Addressing the [Issue](https://github.com/opensearch-project/opensearch-build/issues/3303) 


vars/patchDockerImage.groovy
```diff
@@ -0,0 +1,111 @@
...
+        echo "Trigger docker-scan for ${args.product} version ${inputManifest.build.version}${build_qualifier}"
+        dockerScan: {
+            build job: 'docker-scan',
+            parameters: [
+                string(name: 'IMAGE_FULL_NAME', value: "opensearchstaging/${args.product}:${inputManifest.build.version}${build_qualifier}")
+            ]
+        }
+
+        echo 'Trigger docker-promote'
+        if(args.rerelease){
```
gaiksaya, line 101: Do we want to do few more tests before pushing to Prod? For example we do not want to push to prod if high CVEs are detected in above docker scan. @bbarani @peterzhuamazon 

Also @Divyaasm please add https://github.com/opensearch-project/opensearch-build/blob/main/jenkins/opensearch/distribution-build.jenkinsfile#L440 `wait:true` for each job that is being triggered. I think default is `true` by default but better to mention it.

gaiksaya (changes requested):

vars/patchDockerImage.groovy
```diff
@@ -0,0 +1,111 @@
...
+    set -e
+    set +x
+
+    docker pull ${docker_image}
+    """
+    sh """docker inspect --format '{{ index .Config.Labels "org.label-schema.version"}}' ${docker_image} > versionnumber"""
+
+    sh """docker inspect --format '{{ index .Config.Labels "org.label-schema.build-date"}}' ${docker_image} > time"""
+
+    sh """docker inspect --format '{{ index .Config.Labels "org.label-schema.description"}}' ${docker_image} > number"""
```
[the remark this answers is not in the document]
Divyaasm: Yeah but having all these statements in a single block isn't storing the output value in a variable. That's technically resulting an error.

vars/patchDockerImage.groovy
```diff
@@ -0,0 +1,111 @@
...
+    if (build_qualifier != null && build_qualifier != 'null') {
+        build_qualifier = "-" + build_qualifier
+    }
+    else {
+        build_qualifier = ''
+    }
+
+    if (artifactUrlX64 == null || artifactUrlARM64 ==  null) {
+    echo 'Skipping docker build, one of x64 or arm64 artifacts was not built.'
+    } else {
```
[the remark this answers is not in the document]
Divyaasm: Sure Sayali

vars/patchDockerImage.groovy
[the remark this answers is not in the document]
gaiksaya: That doesn't sound right! This might be the culprit https://github.com/opensearch-project/opensearch-build-libraries/pull/268/files#diff-d42288632ebcbc435313fa808e87e49f58461165639396b6ec30cbb928596336R15-R17

I believe by default the shell is `zsh` and here you are running bash. Also those lines apply only to one command `docker pull ${docker_image}`. Try removing those and put it under default shell

vars/patchDockerImage.groovy
Divyaasm: Think we have ignored CVE's presence in this implementation promote the image when Docker-Scan and Docker-Validation(enhancement) are successful . The idea is to re-run the Job every week to align with Base OS Image. @bbarani Correct me if I'm wrong.

bbarani: I don't think we can do anything about any CVE at this point in time if it didn't remediated as part of `yum update` as the fix needs to come from base OS team.

vars/patchDockerImage.groovy
```diff
@@ -0,0 +1,111 @@
...
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+void call(Map args = [:]) {
+    def lib = library(identifier: 'jenkins@main', retriever: legacySCM(scm))
+    String docker_image = "opensearchproject/${args.product}:${args.tag}"
+
+    sh"""
+    #!/bin/bash
```
peterzhuamazon, line 15: Use this:
```
    sh """#!/bin/bash
```

Your original implementation will not force bash and will default back to sh.

vars/patchDockerImage.groovy
```diff
@@ -0,0 +1,111 @@
...
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+void call(Map args = [:]) {
+    def lib = library(identifier: 'jenkins@main', retriever: legacySCM(scm))
```
peterzhuamazon, line 11: Are we still using lib definition in the lib file? @gaiksaya I thought we have moved away from that for a long time?

vars/patchDockerImage.groovy
```diff
@@ -0,0 +1,111 @@
...
+    String docker_image = "opensearchproject/${args.product}:${args.tag}"
+
+    sh"""
+    #!/bin/bash
+    set -e
+    set +x
+
+    docker pull ${docker_image}
+    """
+    sh """docker inspect --format '{{ index .Config.Labels "org.label-schema.version"}}' ${docker_image} > versionnumber"""
```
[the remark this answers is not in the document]
peterzhuamazon: Use sh("") if you are trying to do one liner unless you have other reasons.

vars/patchDockerImage.groovy
peterzhuamazon, line 25: Why are we keep doing one liners here?

vars/patchDockerImage.groovy
```diff
@@ -0,0 +1,111 @@
...
+    build_time = readFile('time').trim()
+    build_number = readFile('number').trim()
+
+    def inputManifest = lib.jenkins.InputManifest.new(readYaml(file: "manifests/${version}/${args.product}-${version}.yml"))
+
+    artifactUrlX64 = "https://ci.opensearch.org/ci/dbc/distribution-build-${args.product}/${version}/${build_number}/linux/x64/tar/dist/${args.product}/${args.product}-${version}-linux-x64.tar.gz"
+
+    artifactUrlARM64 = "https://ci.opensearch.org/ci/dbc/distribution-build-${args.product}/${version}/${build_number}/linux/arm64/tar/dist/${args.product}/${args.product}-${version}-linux-arm64.tar.gz"
+
+    /*slice the time to get date value*/
```
peterzhuamazon, line 37: Use `// `

vars/patchDockerImage.groovy
```diff
@@ -0,0 +1,111 @@
...
+    build_date = build_time[0..3] + build_time[5..6] + build_time[8..9]
```
peterzhuamazon, line 38: Add example output in the comment.

vars/patchDockerImage.groovy
```diff
@@ -0,0 +1,111 @@
...
+
+    def build_qualifier = inputManifest.build.qualifier
```
peterzhuamazon, line 40: We have discussed to re-use the build lib that runs the exact / similar code.
Are you just copy pasting again them here?
https://github.com/opensearch-project/opensearch-build-libraries/blob/main/vars/buildDockerImage.groovy

peterzhuamazon (changes requested):
Hi @Divyaasm please read the comment here and considering not copy pasting the entire dockerbuild lib if possible.
 https://github.com/opensearch-project/opensearch-build-libraries/blob/main/vars/buildDockerImage.groovy

Thanks.

vars/patchDockerImage.groovy
peterzhuamazon: In our offline discussion the decision was to use this lib while adding new branches of actions if needed.

vars/buildDockerImage.groovy
```diff
@@ -12,6 +12,12 @@ void call(Map args = [:]) {
     def build_qualifier = inputManifest.build.qualifier
     def build_number = args.buildNumber ?: "${BUILD_NUMBER}"
 
+    String image_tag =""
+
+    if (args.buildDate != null){
```
peterzhuamazon, line 17: Also add `!= 'null'` condition as sometimes jenkins does not do pure groovy and can cause weird behavior.

vars/buildDockerImage.groovy
```diff
@@ -12,6 +12,12 @@ void call(Map args = [:]) {
     def build_qualifier = inputManifest.build.qualifier
     def build_number = args.buildNumber ?: "${BUILD_NUMBER}"
 
+    String image_tag =""
```
peterzhuamazon, line 15: Add a space between = and ""

vars/buildDockerImage.groovy
peterzhuamazon, line 17: Since you introduce new build args, can you please update this lib to have all the args and details in comments? 
Thanks.

vars/patchDockerImage.groovy
```diff
@@ -0,0 +1,69 @@
...
+ */
+
+void call(Map args = [:]) {
+    def lib = library(identifier: 'jenkins@main', retriever: legacySCM(scm))
+    String docker_image = "opensearchproject/${args.product}:${args.tag}"
+    boolean tag_latest = false
+
+    if (args.tag == "2"){
+        tag_latest = true
+    }
```
peterzhuamazon, line 17: This is so bad can we please dont hardcode 2 as latest?
We should retrieve the `latest` tag and get the major version instead.
Thanks.

vars/patchDockerImage.groovy
```diff
@@ -0,0 +1,69 @@
...
+        inputManifest: "manifests/${version}/${args.product}-${version}.yml",
+        buildNumber: "${build_number}",
+        buildDate: "${build_date}",
+        buildOption: "${args.rerelease}",
+        artifactUrlX64: "${artifactUrlX64}",
+        artifactUrlArm64: "${artifactUrlARM64}"
+    )
+
+    echo 'Triggering docker-promote'
+    if(args.rerelease == "re_release_docker_image"){
```
peterzhuamazon, line 59: suggest re_release than rerelease to keep the consistency and readability.

peterzhuamazon (commented):
Hi @Divyaasm this is much better and I add some comments on this.
Please take a look when you have time.

Thanks.

vars/buildDockerImage.groovy
Divyaasm: Sure Peter

vars/buildDockerImage.groovy
```diff
@@ -6,11 +6,24 @@
  * this file be licensed under the Apache-2.0 license or a
  * compatible open source license.
  */
+/**
+@param Map[inputManifest] <Required> - path to Input Manifest.
```
peterzhuamazon, line 10: Capital `path`.

vars/patchDockerImage.groovy
```diff
@@ -0,0 +1,77 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
```
peterzhuamazon, line 9: Could you please do the same and add comments for the args?

vars/patchDockerImage.groovy
```diff
@@ -0,0 +1,77 @@
...
+    sh """#!/bin/bash
+    set -e
+    set +x
+    docker pull ${docker_image}
+    docker pull ${latest_docker_image}
+    docker inspect --format '{{ index .Config.Labels "org.label-schema.version"}}' ${docker_image} > versionNumber
+    docker inspect --format '{{ index .Config.Labels "org.label-schema.build-date"}}' ${docker_image} > time
+    docker inspect --format '{{ index .Config.Labels "org.label-schema.description"}}' ${docker_image} > buildNumber
+    docker inspect --format '{{ index .Config.Labels "org.label-schema.version"}}' ${latest_docker_image} > latestVersionNumber
```
peterzhuamazon, line 25: I didnt realize you are creating files here.
You can simply use this to save all the values?
```
GIT_COMMIT_EMAIL = sh (
    script: 'git --no-pager show -s --format=\'%ae\'',
    returnStdout: true
).trim()
```

Possibly I miss it in the initial review?

Thanks.

peterzhuamazon (commented):
Hi @Divyaasm please check more comments here.

Thanks.

vars/patchDockerImage.groovy
gaiksaya: We did not move away from this. Its still being used in most of the lib. The whole version increment workflow is for replacing this `main` to appropriate version when we bump. Also this is required to used lib within lib.

peterzhuamazon (commented):
Great work, thanks.

peterzhuamazon (approved):

gaiksaya (approved):

PR #270, opened 2023-08-11
Add library to create test report manifest

zelinh:
### Description
Add groovy library to trigger report workflow to create test report manifest. Upload the generated `test-report.yml` to the S3 test results bucket. 

### Issues Resolved
Part of https://github.com/opensearch-project/opensearch-build/issues/1274


vars/createUploadTestReportManifest.groovy
```diff
@@ -0,0 +1,98 @@
...
+ @param args.buildManifestDashboards <optional> - The path of the build manifest of OpenSearch Dashboards
+ @param args.testRunID <required> - Test run id of the test workflow being reported.
+ @param args.testType <required> - Type of the test workflow being reported.
+ @param args.componentName <Optional> - Components that workflow runs on.
+ */
+
+def call(Map args = [:]) {
+    lib = library(identifier: 'jenkins@main', retriever: legacySCM(scm))
+
+    def testID = args.testRunID;
```
rishabh6788, line 23: nit: `testId` and `testRunId`.

vars/createUploadTestReportManifest.groovy
```diff
@@ -0,0 +1,98 @@
...
+    def testType = args.testType;
+
+    def testManifest = lib.jenkins.TestManifest.new(readYaml(file: args.testManifest))
+    def buildManifest = lib.jenkins.BuildManifest.new(readYaml(file: args.buildManifest))
+    def buildManifestDashboards = args.buildManifestDashboards ? lib.jenkins.BuildManifest.new(readYaml(file: args.buildManifestDashboards)) : null;
```
rishabh6788, line 28: nit: Does it make sense to rename to `dashboardsBuildManifest`?

vars/createUploadTestReportManifest.groovy
```diff
@@ -0,0 +1,98 @@
...
+    echo "Start Reporting workflow for test type: " + buildManifest.getDistribution()
+
+    String paths = generatePaths(testManifest, buildManifest, buildManifestDashboards)
+    echo "Paths: ${paths}"
+
+    String basePath = generateBasePaths(buildManifest)
+    echo "Base Path ${basePath}"
+
+    String component = args.componentName
+    echo "Component: ${component}"
```
gaiksaya, line 39: Do we need to log all of this stuff?

vars/createUploadTestReportManifest.groovy
```diff
@@ -0,0 +1,98 @@
...
+String generateBasePaths(buildManifest) {
+    return ["${env.PUBLIC_ARTIFACT_URL}", "${env.JOB_NAME}", buildManifest.build.version, buildManifest.build.id, buildManifest.build.platform, buildManifest.build.architecture, buildManifest.build.distribution].join("/")
+}
+
+String generateUploadPath(testManifest, buildManifest, buildManifestDashboards, testID) {
+    def product = testManifest.name
+    def productBuildManifest = (product.equals("OpenSearch")) ? buildManifest : buildManifestDashboards
+
+    String buildId = productBuildManifest.build.id
+    echo "Build Id: ${buildId}"
```
gaiksaya, line 94: Same as above

tests/jenkins/TestCreateUploadTestReportManifest.groovy
```diff
@@ -0,0 +1,46 @@
...
+                'integ-test',
+                '',
+            )
+        )
+        super.setUp()
+        super.testPipeline("tests/jenkins/jobs/CreateUploadTestReportManifest_Jenkinsfile")
+        assertThat(getShellCommands('sh', 'report.sh'), hasItems('./report.sh tests/data/opensearch-1.3.0-test.yml --artifact-paths opensearch=https://ci.opensearch.org/ci/dbc/distribution-build-opensearch/1.3.0/c3ff7a232d25403fa8cc14c97799c323/linux/x64/tar --test-run-id 1234 --test-type integ-test --base-path null/null/1.3.0/c3ff7a232d25403fa8cc14c97799c323/linux/x64/tar --component null'))
+    }
+
+    def getShellCommands(methodName, searchString) {
```
gaiksaya, line 36: Can you add assertions for dashboards as well? To make sure OS is getting picked up?

gaiksaya (changes requested):

vars/createUploadTestReportManifest.groovy
zelinh: I think logging them could help us looking for issues if anything broken, WDYT?

vars/createUploadTestReportManifest.groovy
```diff
@@ -0,0 +1,98 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+/** Library to close GitHub issue across opensearch-project repositories.
```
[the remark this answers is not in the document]
zelinh: Thanks!

vars/createUploadTestReportManifest.groovy
```diff
@@ -0,0 +1,98 @@
...
+
+    String reportCommand =
+            [
+                    './report.sh',
+                    "${args.testManifest}",
+                    "--artifact-paths ${paths}",
+                    "--test-run-id ${testID}",
+                    "--test-type ${testType}",
+                    "--base-path ${basePath}",
+                    "--component ${component}",
```
rishabh6788, line 49: Since the component is an optional parameter, you can have a null check like [here](https://github.com/opensearch-project/opensearch-build-libraries/blob/main/vars/runBenchmarkTestScript.groovy#L106) and then change it to 
`isNullOrEmpty(args.componentName.toString()) ? "" : "--component ${args.componentName}",`

zelinh: Good suggestion. Updating here. Thanks!

vars/createUploadTestReportManifest.groovy
```diff
@@ -0,0 +1,98 @@
+/*
```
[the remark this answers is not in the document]
zelinh: I honestly prefer the current one since we haven't add `and` between actions we were doing. e.g. `buildAssembleUpload.groovy`.

tests/jenkins/TestCreateUploadTestReportManifest.groovy
zelinh: Updated.

vars/createUploadTestReportManifest.groovy
zelinh: Updated.

vars/createUploadTestReportManifest.groovy
zelinh: Makes sense to me. Updated the name.

zelinh:
Just had a new commit to accommodate your suggestions. Please help take a look again. Thanks! @gaiksaya @rishabh6788

tests/jenkins/TestCreateUploadTestReportManifest.groovy
```diff
@@ -0,0 +1,47 @@
...
+                'tests/data/opensearch-dashboards-build-1.3.0.yml',
+                '1234',
+                'integ-test',
+                '',
+            )
+        )
+        super.setUp()
+        super.testPipeline("tests/jenkins/jobs/CreateUploadTestReportManifest_Jenkinsfile")
+        assertThat(getShellCommands('sh', 'report.sh'), hasItems('./report.sh tests/data/opensearch-1.3.0-test.yml --artifact-paths opensearch=https://ci.opensearch.org/ci/dbc/distribution-build-opensearch/1.3.0/c3ff7a232d25403fa8cc14c97799c323/linux/x64/tar --test-run-id 1234 --test-type integ-test --base-path null/null/1.3.0/c3ff7a232d25403fa8cc14c97799c323/linux/x64/tar '))
+        assertThat(getShellCommands('sh', 'report.sh'), hasItems('./report.sh tests/data/opensearch-1.3.0-test.yml --artifact-paths opensearch=https://ci.opensearch.org/ci/dbc/distribution-build-opensearch/1.3.0/c3ff7a232d25403fa8cc14c97799c323/linux/x64/tar opensearch-dashboards=https://ci.opensearch.org/ci/dbc/distribution-build-opensearch-dashboards/1.3.0/25b38c278cdd45efa583765d8ba76346/linux/x64/tar --test-run-id 1234 --test-type integ-test --base-path null/null/1.3.0/c3ff7a232d25403fa8cc14c97799c323/linux/x64/tar '))
```
gaiksaya, line 34: This should have opensearch-dashboards test manifest. Also base path is going null here: 
```--base-path null/null/1.3.0/c3ff7a232d25403fa8cc14c97799c323/linux/x64/tar```

File: tests/jenkins/lib-testers/CreateUploadTestReportManifestLibTest.groovy [hunk 10]
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    void configure(helper, binding) {
+        binding.setVariable('env', ['JOB_NAME': 'dummy_integ_test'])
+        binding.setVariable('ARTIFACT_BUCKET_NAME', 'DUMMY_BUCKET_NAME')
+        binding.setVariable('AWS_ACCOUNT_PUBLIC', 'DUMMY_AWS_ACCOUNT_PUBLIC')
+        binding.setVariable('ARTIFACT_BUCKET_NAME', 'DUMMY_ARTIFACT_BUCKET_NAME')
+        binding.setVariable('PUBLIC_ARTIFACT_URL', 'DUMMY_PUBLIC_ARTIFACT_URL')
+        binding.setVariable('env', ['BUILD_NUMBER': '487'])
+        binding.setVariable('STAGE_NAME', 'DUMMY_STAGE_NAME')
```
gaiksaya, line 39: None of these variables are used in this library. Can you check and remove the unsed ones?
Thanks!

vars/createUploadTestReportManifest.groovy
gaiksaya: I believe it adds too much of information. If this would have been a logger we would have put this in debug mode. I don't think they prove to be much useful tbh. Your call.

vars/createUploadTestReportManifest.groovy
```diff
@@ -0,0 +1,100 @@
...
+ @param args.dashboardsBuildManifest <optional> - The path of the build manifest of OpenSearch Dashboards
+ @param args.testRunID <required> - Test run id of the test workflow being reported.
+ @param args.testType <required> - Type of the test workflow being reported.
+ @param args.componentName <Optional> - Components that workflow runs on.
+ */
+
+def call(Map args = [:]) {
+    lib = library(identifier: 'jenkins@main', retriever: legacySCM(scm))
+
+    def testRunID = args.testRunID;
```
gaiksaya, line 23: Do you want to add parameterCheck function to get all the required params? Similar to https://github.com/opensearch-project/opensearch-build-libraries/blob/99c487d8be8540252bd73818a89d311711144b3e/vars/publishToNpm.groovy#L33-L47

zelinh: I believe I'm asserting and verifying those required parameter in the libtester to make sure they are not null.

gaiksaya: That won't verify on actual jenkins end. Lib tester is only for testing the library not actual checks.

zelinh: Oh I see your point here. Do you think it's that necessary? We would be passing these required parameters for sure in the integ test jenkinsfile instead of passing by ourself manually. Even the worst case that some of those parameters are not present, i don't think we want to fail the testing because of this reporting issue and then we have to run the tests again. The only error would only appear on the test-report manifest if that's the case.

tests/jenkins/lib-testers/CreateUploadTestReportManifestLibTest.groovy
zelinh: Yeah you are right! and also explains those `null` in the `base-path`. Updated in the latest commit.

tests/jenkins/TestCreateUploadTestReportManifest.groovy
zelinh: Yeah. Updated here with a new OSD test manifest for 1.3.0.

vars/createUploadTestReportManifest.groovy
gaiksaya: Its the opposite. This will only fail when we try to create the report and not the entire workflow. From library's perspective you can add what args are hard core required. I know we didn't do that before but this helps is fail fast and debug easier than figuring out in python code what went wrong.

zelinh: We can have an offline discussion on this. I may not get your points here since I think it might be not that helpful in this case.

zelinh: Had an offline discussion with @gaiksaya, since we are on pace to make all library with parameter checks, adding checks here as well. Please help review as the checks are added now. If the required parameter is null or empty, we won't fail the job but skip the rest of the workflow.

vars/createUploadTestReportManifest.groovy
```diff
@@ -0,0 +1,122 @@
...
+ @param args.dashboardsBuildManifest <optional> - The path of the build manifest of OpenSearch Dashboards
+ @param args.testRunID <required> - Test run id of the test workflow being reported.
+ @param args.testType <required> - Type of the test workflow being reported.
+ @param args.componentName <Optional> - Components that workflow runs on.
+ */
+
+def call(Map args = [:]) {
+    lib = library(identifier: 'jenkins@main', retriever: legacySCM(scm))
+
+    if (!parameterCheck(args.testManifest, args.buildManifest, args.testRunID, args.testType)) return
```
gaiksaya, line 23: What is this returning?

zelinh: It's not returning anything. It's returned from this library and skips the rest of code below.

gaiksaya, line 23:
```suggestion
    if (!parameterCheck(args.testManifest, args.buildManifest, args.testRunID, args.testType)) return null
```

zelinh: Updated as suggested.

gaiksaya (approved):
Please bump the version to next minor.
Thanks!

PR #302, opened 2023-09-06
[NodeJS18 Upgrade] Add AL2023 Docker Images Selection by BuildDocker lib

peterzhuamazon:
### Description
[NodeJS18 Upgrade] Add AL2023 Docker Images Selection by BuildDocker lib

### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/1563
https://github.com/opensearch-project/opensearch-build/issues/3743

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -120,7 +120,7 @@ jacocoTestReport {
     }
 }
 
-String version = '5.7.2'
+String version = '5.8.0'
 
 task updateVersion {
     doLast {
--- a/build.gradle
+++ b/build.gradle
@@ -120,7 +120,7 @@ jacocoTestReport {
     }
 }
 
-String version = '5.7.1'
+String version = '5.7.2'
 
 task updateVersion {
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -120,7 +120,7 @@ jacocoTestReport {
     }
 }
 
-String version = '5.7.1'
+String version = '5.7.2'
```
gaiksaya, line 123: If this is a breaking change the bump should be 6.0.0. If not then 5.8.0.
Patch releases are only for CVEs or bug fixes. This does not look like a bug but enhancement.
Thanks!

gaiksaya (changes requested):

build.gradle
peterzhuamazon: It is not breaking I will just bump to 5.8.0 then.
Thanks.

gaiksaya (approved):

Issue #320, opened 2021-11-16
Add source code linting to Groovy code and Jenkinsfile(s)


dblock:
### Is your feature request related to a problem? Please describe

Add source code linting to Groovy code and Jenkinsfile(s)

### Describe the solution you'd like

Code in src/jenkins and tests/jenkins, along with Jenkinsfile in jenkins/*.

- Documentation on how to lint in DEVELOPER_GUIDE.
- Pre-commit hook.
- Linting that fails CI if the code is not properly formatted.

chawinphat:
I found this https://github.com/nvuillam/npm-groovy-lint/blob/master/README.md which seems to work quite well. The problem is that the build-libraries repository does not have npm installed.

We can also use jenkins pipeline development tools to lint jenkinsfiles https://www.jenkins.io/doc/book/pipeline/development/.

PR #325, opened 2023-10-15
Added source code linting to Groovy code and Jenkinsfile

chawinphat:
### Description
_Describe what this change achieves._
Created a pre-commit hook to run a linter for groovy and jenkinsfiles.

### Issues Resolved
[#320](https://github.com/opensearch-project/opensearch-build-libraries/issues/320)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,14 @@
+---
+default_stages: [commit]
+repos:
+  - repo: local
+    hooks:
+      - id: npm-groovy-lint
+        stages: [commit]
+        name: npm-groovy-lint
+        entry: bash -c 'npm-groovy-lint tests/jenkins'
+        language: system
+        types: [groovy]
+        description: Groovy & Jenkinsfile Linter
+  
+  
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -10,5 +10,3 @@ repos:
         language: system
         types: [groovy]
         description: Groovy & Jenkinsfile Linter
-  
-  
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,7 +6,7 @@ repos:
       - id: npm-groovy-lint
         stages: [commit]
         name: npm-groovy-lint
-        entry: bash -c 'npm-groovy-lint tests/jenkins'
+        entry: bash -c 'npm-groovy-lint tests/jenkins src/jenkins'
         language: system
         types: [groovy]
         description: Groovy & Jenkinsfile Linter


gaiksaya:
Hi @chawinphat 

Thank you for the contribution. We also have (most of the) groovy code living in `vars` folder. Please include that too.
Also can you elaborate a bit on how this is supposed to work? Maybe adding a working snippet?
Thanks!

.groovylintrc.json
```diff
@@ -1,5 +1,5 @@
 {
-    "extends": "recommended",
+    "extends": "recommended-jenkinsfile",
```
gaiksaya, line 2: What is the difference between recommended and recommended-jenkinsfile?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,12 @@
+---
+default_stages: [commit]
+repos:
+  - repo: local
+    hooks:
+      - id: npm-groovy-lint
+        stages: [commit]
+        name: npm-groovy-lint
+        entry: bash -c 'npm-groovy-lint tests/jenkins src/jenkins vars --failon error'
```
gaiksaya, line 9: What does a failure look like? Will it won't even allow to commit the code or commit and then error out?

gaiksaya (commented):
Thanks for contributing this feature @chawinphat .
Couple of questions mentioned below. Overall looks good.

.groovylintrc.json
chawinphat: It's the linter's recommended (format)[https://github.com/nvuillam/npm-groovy-lint/tree/main#format] which lints jenkinsfiles too. When using just 'recommended' the linter catches errors that are not actually errors and breaks the program.

.pre-commit-config.yaml
chawinphat: It won't allow you to commit code. Errors only occur when an error is detected and this is set by the --failon flag.

.groovylintrc.json
gaiksaya: Can you provide some examples what failures are you encountering? I am doubtful what to use as well. On high level looks like `recommended` is best path for groovy. `recommended-jenkinsfile` seems to be applicable for jenkinsFiles specifically? Please correct me if I am wrong. 

Adding @prudhvigodithi @peterzhuamazon @dblock to get some input.

chawinphat: One error I saw was that 'variable x was not used in the file'. However, x was referenced inside a string. 

Running the linter's automatic fixing caused tests to fail if I used the 'jenkins' setting while it was able to build and test fine with 'jenkins-file'

PR #326, opened 2023-10-16
Add support to run ad-hoc benchmark using distribution url

rishabh6788:
### Description
This PR:
* Adds support to run benchmark using distribution url. Related to https://github.com/opensearch-project/opensearch-build/pull/4098
* Adds additional sleep to avoid pip install conflict errors when run with and without security in-parallel. 

### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/4073. 


vars/runBenchmarkTestScript.groovy
```diff
@@ -46,9 +52,14 @@ void call(Map args = [:]) {
...
-            happening in parallel to download resulting in file not found error
+            happening in parallel to download resulting in file not found error. To avoid pip install conflict errors
+            when runnin with and without security run in parallel add enough gap between execution.
              */
-            sleep(5)
+            if (args.insecure.toBoolean()) {
+                sleep(5)
+            } else {
+                sleep(15)
+            }
```
peterzhuamazon, line 62: It is possible that two insecure or secure runs happening on the same agent.
Would recommend to randomize the wait time between 10-30 seconds.

Or any other way to achieve this? Thanks.

rishabh6788: At the moment only two runs can happen on a single agent (with and without-security). 
We can start with this to see if it has any effect on the runs.

peterzhuamazon: There are many other scenarios where agents will end up scheduling runs with same spec, making this change not as efficient as it seems.

rishabh6788: The probability of this happening is highly unlikely since we are using `H` to add jitter between two scheduled runs so that jenkins does not start multiple runs at the same time. 
Also, using random also does not guarantee that at certain point it throws up the same random numbers to two independent runs.

peterzhuamazon: @rishabh6788 does sec and nosec run trigger in parallel now for each run?

rishabh6788: yes, they are triggered in parallel for each jenkins execution.

peterzhuamazon (approved):

PR #357, opened 2023-12-18
Pass in initial admin password in jenkins verification library

derek-ho:
### Description
Starting in 2.12.0, installing the security demo configuration requires an initial admin password to be passed in. This changes some validation library steps to expect this. 

### Issues Resolved

### DIFF
--- a/vars/rpmDashboardsDistValidation.groovy
+++ b/vars/rpmDashboardsDistValidation.groovy
@@ -58,7 +58,7 @@ def call(Map args = [:]) {
 
     //Start the installed OpenSearch-Dashboards distribution
     systemdCommands(
-            command: "sudo env OPENSEARCH_INITIAL_ADMIN_PASSWORD=myStrongPassword123! start",
+            command: "start",
             product: "opensearch"
     )
     systemdCommands(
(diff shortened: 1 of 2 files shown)


derek-ho:
Related to: https://github.com/opensearch-project/opensearch-build/pull/4302

derek-ho:
@peterzhuamazon @rishabh6788 @gaiksaya would this run into the same issue? Would < 2.12.0 versions of OpenSearch be built using this validation library and fail because the admin:admin changes are not consumed in those prior versions?

vars/rpmDashboardsDistValidation.groovy
```diff
@@ -58,7 +58,7 @@ def call(Map args = [:]) {
 
     //Start the installed OpenSearch-Dashboards distribution
     systemdCommands(
-            command: "start",
+            command: "sudo env OPENSEARCH_INITIAL_ADMIN_PASSWORD=myStrongPassword123! start",
```
peterzhuamazon, line 61: This will result in `sudo systemctl sudo env OPENSEARCH_INITIAL_ADMIN_PASSWORD=myStrongPassword123! start`

vars/rpmOpenSearchDistValidation.groovy
```diff
@@ -46,7 +46,7 @@ def call(Map args = [:]) {
     //Install OpenSearch with designated version via yum
     println("Start installation with yum.")
     rpmCommands(
-            command: "install",
+            command: "sudo env OPENSEARCH_INITIAL_ADMIN_PASSWORD=myStrongPassword123! install",
```
peterzhuamazon, line 49: Are you expecting user to first define the env var before running `yum/rpm install`?

peterzhuamazon (changes requested):

vars/rpmOpenSearchDistValidation.groovy
peterzhuamazon: You need to change the spec files of both deb and rpm, check if the var is set else exit with error msgs in the console.

DarshitChanpura:
this should be unblocked by: https://github.com/opensearch-project/opensearch-build/pull/4332

vars/rpmDashboardsDistValidation.groovy
DarshitChanpura: don't think we need to pass variable here

vars/rpmDashboardsDistValidation.groovy
```diff
@@ -84,12 +84,12 @@ def call(Map args = [:]) {
         if (osd_status_json != 0) {
             sleep 10
             osd_status_json = sh (
-                    script: "curl -s \"http://localhost:5601/api/status\" -u admin:admin",
+                    script: "curl -s \"http://localhost:5601/api/status\" -u admin:myStrongPassword123!",
```
DarshitChanpura, line 87: this change should be version based. (admin for < 2.12 and myStrongPassword123! >= 2.12)

derek-ho:
Addressed all comments. @peterzhuamazon can you take a look, pending the companion PR in the build repo? Thanks!

vars/rpmOpenSearchDistValidation.groovy
derek-ho: Addressed in build repo companion PR.

derek-ho:
@peterzhuamazon @prudhvigodithi is this PR even needed anymore? I am not sure how this library is being used, if at all during the build process.

peterzhuamazon:
Close this as it will be replaced as part of the validation workflow.

cc: @zelinh @Divyaasm 

Thanks.

PR #362, opened 2024-01-13
Add a new groovy library to download the previous build artifacts

zelinh:
### Description
Add a new groovy library to download the previous build artifacts intended to be used for incremental build.

### Issues Resolved
Part of https://github.com/opensearch-project/opensearch-build/issues/4292


vars/buildManifest.groovy
```diff
@@ -28,5 +35,6 @@ void call(Map args = [:]) {
         args.snapshot ? '--snapshot' : null,
         args.lock ? '--lock' : null,
         args.continueOnError ? '--continue-on-error' : null,
+        incremental_enabled ? '--incremental': null
```
peterzhuamazon, line 38: Add a space after the `'--incremental'` param.

vars/retrievePreviousBuild.groovy
```diff
@@ -0,0 +1,67 @@
...
+ @param args.platform <required> - Platform of previous build to retrieve.
+ @param args.architecture <required> - Architecture of previous build to retrieve.
+ @param args.distribution <required> - Distribution of previous build to retrieve.
+ @param args.incremental <required> - String value contains the build number of previous build or the latest.
+ */
+void call(Map args = [:]) {
+    def lib = library(identifier: 'jenkins@jenkins_incremental', retriever: legacySCM(scm))
+    def inputManifestObj = lib.jenkins.InputManifest.new(readYaml(file: args.inputManifest))
+
+    def DISTRIBUTION_JOB_NAME = "${JOB_NAME}"
```
peterzhuamazon, line 21: Do you need to ensure it is the correct job name?

peterzhuamazon: Or just use the lib call to parse as arg?

vars/retrievePreviousBuild.groovy
```diff
@@ -0,0 +1,67 @@
...
+    def revision = inputManifestObj.build.version
+
+    def DISTRIBUTION_PLATFORM = args.platform
+    def DISTRIBUTION_ARCHITECTURE = args.architecture
+    def distribution = args.distribution
+    def prefixPath = "${WORKSPACE}/download"
+    def DISTRIBUTION_BUILD_NUMBER
+
+    if (args.incremental.equalsIgnoreCase("latest")) {
```
peterzhuamazon, line 30: Can this arg `incremental` being changed to `version`? As the value is either the version or `latest`.

vars/retrievePreviousBuild.groovy
```diff
@@ -0,0 +1,67 @@
...
+        withCredentials([string(credentialsId: 'jenkins-artifact-bucket-name', variable: 'ARTIFACT_BUCKET_NAME')]) {
+            downloadFromS3(
+                    assumedRoleName: "opensearch-bundle",
+                    roleAccountNumberCred: "jenkins-aws-account-public",
+                    downloadPath: "${DISTRIBUTION_JOB_NAME}/${revision}/index.json",
+                    bucketName: "${ARTIFACT_BUCKET_NAME}",
+                    localPath: "${prefixPath}/index.json",
+                    force: true,
+            )
+        }
```
peterzhuamazon, line 40: I didnt realize this is just a index.json, then curl with `|` would do the trick instantly.
I would agree if it is the full ledge distribution folder, s3 is sufficient.
But for a single file and you only need to parse the value, wouldnt http is easier to go?

vars/retrievePreviousBuild.groovy
```diff
@@ -0,0 +1,67 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+/** Library to call build.sh script (builds OpenSearch and OpenSearch Dashboards distribution)
```
gaiksaya, line 9: Fix the lib definition? I believe this lib only downloads the  previous builds?

vars/buildManifest.groovy
```diff
@@ -18,6 +18,13 @@
  @param args.continueOnError <optional> - Do not fail the distribution build on any plugin component failure. Defaults to null
  */
 void call(Map args = [:]) {
+    boolean incremental_enabled = args.incremental != null && !args.incremental.isEmpty() && !args.incremental.equalsIgnoreCase("false")
+
+    if (incremental_enabled) {
+        echo("Start retreiving library")
```
gaiksaya, line 24:
```suggestion
        echo("Incremental build enabled! Retrieving previous build library")
```

vars/retrievePreviousBuild.groovy
gaiksaya: or maybe be `incrementally_build_id`

vars/retrievePreviousBuild.groovy
zelinh: Yes that's true. However, the URL for downloading the `index.json` from beta jenkins for testing is not available. That's the reason I used s3Download to be compatible for both.

vars/retrievePreviousBuild.groovy
```diff
@@ -0,0 +1,67 @@
...
+        downloadFromS3(
+                assumedRoleName: "opensearch-bundle",
+                roleAccountNumberCred: "jenkins-aws-account-public",
+                downloadPath: "${artifactPath}/",
+                bucketName: "${ARTIFACT_BUCKET_NAME}",
+                localPath: "${prefixPath}",
+                force: true,
+        )
+    }
+    sh("mkdir -p ${distribution} && mv ${prefixPath}/${artifactPath} ${WORKSPACE}")
```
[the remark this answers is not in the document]
zelinh: Updated.

vars/retrievePreviousBuild.groovy
gaiksaya: I think we can change to curl or wget. Credentials should not be required if the file path is known.

gaiksaya: Oh right! Downloading directly from S3 would cause issues. Better to use curl/wget to get index.json. @Divyaasm Can add more.
Thanks!

vars/retrievePreviousBuild.groovy
zelinh: But this is the argument we would get from `buildManifest` groovy library. It could also be `false` there not just a simple id.

vars/retrievePreviousBuild.groovy
zelinh: Updated.

vars/buildManifest.groovy
zelinh: Updated. Thanks!

vars/retrievePreviousBuild.groovy
zelinh: Just sync up with @Divyaasm, we will be having different paths for index.json and it's unique for different type of distribution/architecture/platform. 

For now, I can use `curl` to download from current location for index.json 
e.g. `https://ci.opensearch.org/ci/dbc/${DISTRIBUTION_JOB_NAME}/${revision}/index.json`

But in the future once we have new `index.json` for each distribution/architecture/platform type, we would need to update the URL here to 
`https://ci.opensearch.org/ci/dbc/${DISTRIBUTION_JOB_NAME}/${revision}/index/${platform}/${architecture}/${distribution}/index.json`

What do you think? @gaiksaya @peterzhuamazon

gaiksaya: I think it should be always curl. downloadfromS3 is over complicating a single file download.

zelinh: Sure. I have to use s3download since we don't have that URL available for beta Jenkins. But in here, we can use curl

zelinh: Updated in the latest commit. Please help take a look. @gaiksaya @peterzhuamazon

vars/retrievePreviousBuild.groovy
gaiksaya: Hi @zelinh I still don't see the updated definition of the library. Was this missed?

vars/retrievePreviousBuild.groovy
gaiksaya: Since empty/null/false are all same in this case, `previous_build_id` seems good fit. How about replacing `false` with `none/null` here? or @peterzhuamazon may have better ideas?

vars/retrievePreviousBuild.groovy
zelinh: Seems like I updated at wrong place. Just made a new commit on this.

vars/retrievePreviousBuild.groovy
zelinh: I would like to keep it as the original `incremental` name for the argument. 
1. This argument will be initiated in the jenkinsfile and incremental is straightforward stating what this parameter is for.
2. This parameter will work as a dynamic option for user: enable/disable the incremental build process; if enable, give user option to choose what particular build artifacts to download. Since using Jenkins dynamic parameter is a little complicated, I lean on to name it more generic. 
What do you think? @gaiksaya @peterzhuamazon

gaiksaya: You can add that customization in main distribution file. When it comes to this specific library, you need a previous build id which needs to be mentioned explicitly. Still leaning toward `previous_build_id` for this specific library. How you showcase that at the distribution pipeline level can be different that this.

vars/retrievePreviousBuild.groovy
gaiksaya: This will take any job name where this is run? If someone wants runs this lib in another job, this might cause issue. Take it as an optional arg and default to ${JOB_NAME}

zelinh: @gaiksaya Done in the latest commit.

vars/retrievePreviousBuild.groovy
zelinh: Updated the PR with two separate parameters. One is a boolean for whether we enable incremental build; this will be used in `buildManifest` lib to determine whether trigger incremental build. The other is the generic one build ID to download previous build used in `retrievePreviousBuild` library. Please take a look. Thanks. @gaiksaya @peterzhuamazon

vars/buildManifest.groovy
```diff
@@ -16,8 +16,16 @@
...
  @param args.lock <optional> - Generate a stable reference manifest. Defaults to null.
  @param args.continueOnError <optional> - Do not fail the distribution build on any plugin component failure. Defaults to null
+ @param args.incremental <optional> - Boolean value to enable incremental build.
  */
 void call(Map args = [:]) {
+    boolean incremental_enabled = args.incremental != null && args.incremental
+
+    if (incremental_enabled) {
+        echo("Incremental build enabled! Retrieving previous build library.")
+        retrievePreviousBuild(args)
```
gaiksaya, line 26: The arg `previousBuildId` is not passed here.
Is the default `latest` here?

vars/retrievePreviousBuild.groovy
```diff
@@ -0,0 +1,59 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+/** Library to download previous build artifacts used for incremental build.
```
gaiksaya, line 9: nit:

```suggestion
/** Library to download previous build artifacts into the workspace.
```

vars/retrievePreviousBuild.groovy
```diff
@@ -0,0 +1,59 @@
...
+ @param args.architecture <required> - Architecture of previous build to retrieve.
+ @param args.distribution <required> - Distribution of previous build to retrieve.
+ @param args.previousBuildId <optional> - Build id of previous build for incremental build. Defaults to latest.
+ */
+void call(Map args = [:]) {
+    def lib = library(identifier: 'jenkins@main', retriever: legacySCM(scm))
+    def inputManifestObj = lib.jenkins.InputManifest.new(readYaml(file: args.inputManifest))
+
+    def DISTRIBUTION_JOB_NAME = args.jobName ? args.jobName : "${JOB_NAME}"
+    def revision = inputManifestObj.build.version
```
gaiksaya, line 23: Should this be named `version`?

vars/retrievePreviousBuild.groovy
```diff
@@ -0,0 +1,59 @@
...
+ * compatible open source license.
+ */
+/** Library to download previous build artifacts used for incremental build.
+ @param Map args = [:] args A map of the following parameters
+ @param args.inputManifest <required> - Relative path to input manifest containing all the components to build.
+ @param args.jobName <optional> - Relative path to input manifest containing all the components to build.
+ @param args.platform <required> - Platform of previous build to retrieve.
+ @param args.architecture <required> - Architecture of previous build to retrieve.
+ @param args.distribution <required> - Distribution of previous build to retrieve.
+ @param args.previousBuildId <optional> - Build id of previous build for incremental build. Defaults to latest.
```
gaiksaya, line 16: I think you forgot to assign the default value. if the value is null or empty please assign `previousBuildId = latest`

zelinh: This default value will not be assigned here right? I think the default should be assigned at Job level when we initiate the parameter.

vars/retrievePreviousBuild.groovy
zelinh: Updated.

vars/retrievePreviousBuild.groovy
zelinh: Renamed. Thanks.

vars/buildManifest.groovy
zelinh: It will be passed from the arguments of this class right? The default latest is assigned when we declare the parameter.

vars/retrievePreviousBuild.groovy
zelinh: Had an offline discussion with @gaiksaya. Just updated a new line for defaulting this `previousBuildId` to `latest` if user doesn't provide argument.

vars/buildManifest.groovy
gaiksaya: Since the default is pointing to latest we should be good but if not, whoever calls this lib needs to provide `args.previousBuildId`. Not sure what is the best approach. Maybe mentioning in the comments above?

gaiksaya (approved):
@peterzhuamazon Maybe take another look?

peterzhuamazon (approved):

PR #377, opened 2024-02-01
Add library to add and remove GH issue labels

gaiksaya:
### Description
 Add library to add and remove GH issue labels

### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/4358


vars/updateGitHubIssueLabels.groovy
```diff
@@ -0,0 +1,101 @@
...
+def addAction(args, issueNumber) {
+    List<String> allLabels = Arrays.asList(args.label.split(','))
+    allLabels.each { i ->
+        try {
+            def name = sh(
+                    script: "gh label list --repo ${args.repoUrl} -S ${i} --json name --jq '.[0].name'",
+                    returnStdout: true
+                ).trim()
+            if (name.equals(i.trim())) {
+                println("Label ${i} already exists. Adding it to the issue")
```
prudhvigodithi, line 54: @gaiksaya do we need another if condition to check if the label is already added to the issue and skip if added? I'm fine if github does not throw an error when running `add` operation even though label exists but just could save an API.

vars/updateGitHubIssueLabels.groovy
```diff
@@ -0,0 +1,101 @@
...
+    allLabels.each { i ->
+        try {
+            def name = sh(
+                    script: "gh label list --repo ${args.repoUrl} -S ${i} --json name --jq '.[0].name'",
+                    returnStdout: true
+                ).trim()
+            if (name.equals(i.trim())) {
+                println("Removing label ${i} from the issue")
+                sh(
+                    script: "gh issue edit ${issueNumber} -R ${args.repoUrl} --remove-label \"${i}\"",
```
prudhvigodithi, line 84: Same as above, check if does not exist and skip the `gh issue edit`.

vars/updateGitHubIssueLabels.groovy
gaiksaya: Right! It does not throw an error on already existing condition thats why I chose to ignore it. Do not want to waste more API calls for checks.

prudhvigodithi (approved):

PR #380, opened 2024-02-02
Combine create and close GH issues for build failures into one library

gaiksaya:
### Description
Combine create and close GH issues for build failures into one library.

TL;DR:
The GH issues won't be blindly closed if one of the platform passes the build in case of distribution failures. It will only be closed if all platforms have passed for given component.

### Issues Resolved
resolves https://github.com/opensearch-project/opensearch-build/issues/4357

### DIFF
--- a/vars/UpdateBuildFailureIssues.groovy
+++ b/vars/UpdateBuildFailureIssues.groovy
@@ -16,8 +16,8 @@ void call(Map args = [:]) {
     lib = library(identifier: 'jenkins@main', retriever: legacySCM(scm))
     def failureMessages = args.failureMessages
     def passMessages = args.passMessages
-    def yamlFile = readYaml(file: args.inputManifestPath)
-    def currentVersion = yamlFile.build.version
+    def inputManifest = readYaml(file: args.inputManifestPath)
+    def currentVersion = inputManifest.build.version
 
     List<String> failedComponents = []
     List<String> passedComponents = []
@@ -35,7 +35,7 @@ void call(Map args = [:]) {
     }
     passedComponents = passedComponents.unique()
 
-    for (component in yamlFile.components) {
+    for (component in inputManifest.components) {
         if (failedComponents.contains(component.name)) {
             println("Component ${component.name} failed, creating github issue")
             exactComponentFailureMessage = getExactErrorMessage(failureMessages, component.name)
@@ -48,8 +48,7 @@ void call(Map args = [:]) {
                 issueTitle: "[AUTOCUT] Distribution Build Failed for ${component.name}-${currentVersion}",
                 issueBody: ghIssueBody,
                 label: "autocut,v${currentVersion}"
-                )
-            sleep(time:3, unit:'SECONDS')
+            )
         } else if (passedComponents.contains(component.name) && !failedComponents.contains(component.name)) {
             println("Component ${component.name} passed, closing github issue")
             ghIssueBody = """Closing the issue as the distribution build for ${component.name} has passed for version: **${currentVersion}**.
@@ -60,8 +59,8 @@ void call(Map args = [:]) {
                 closeComment: ghIssueBody,
                 label: "autocut,v${currentVersion}"
             )
-            sleep(time:3, unit:'SECONDS')
         }
+        sleep(time:3, unit:'SECONDS')
     }
 }
 
--- a/vars/UpdateBuildFailureIssues.groovy
+++ b/vars/UpdateBuildFailureIssues.groovy
@@ -1,0 +1,74 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+ /** Library to create, update an close GitHub issue across opensearch-project repositories for distribution build failures.
+ @param Map args = [:] args A map of the following parameters
+ @param args.failureMessages <required> - Failure message retrieved from buildFailureMessage() method.
+ @param args.passMessages <required> - Passing message retrieved from buildFailureMessage() method.
+ @param args.inputManifestPath <required> - Path to input manifest.
+ */
+void call(Map args = [:]) {
+    lib = library(identifier: 'jenkins@main', retriever: legacySCM(scm))
+    def failureMessages = args.failureMessages
+    def passMessages = args.passMessages
+    def yamlFile = readYaml(file: args.inputManifestPath)
+    def currentVersion = yamlFile.build.version
+
+    List<String> failedComponents = []
+    List<String> passedComponents = []
+
+    for (message in failureMessages) {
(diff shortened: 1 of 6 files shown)


vars/UpdateBuildFailureIssues.groovy
```diff
@@ -0,0 +1,74 @@
...
+    List<String> passedComponents = []
+
+    for (message in failureMessages) {
+        java.util.regex.Matcher match = (message =~ /(?<=\bError building\s).*/)
+        String matched = match[0]
+        failedComponents.add(matched.split(' ')[0].split(',')[0].trim())
+    }
+    failedComponents = failedComponents.unique()
+    for (message in passMessages) {
+        java.util.regex.Matcher match = (message =~ /(?<=\bSuccessfully built\s).*/)
```
peterzhuamazon, line 32: Are these two msg reliable and unique enough to decide the success and failure of a component? Just wonderin.

vars/UpdateBuildFailureIssues.groovy
```diff
@@ -0,0 +1,74 @@
...
+ /** Library to create, update an close GitHub issue across opensearch-project repositories for distribution build failures.
+ @param Map args = [:] args A map of the following parameters
+ @param args.failureMessages <required> - Failure message retrieved from buildFailureMessage() method.
+ @param args.passMessages <required> - Passing message retrieved from buildFailureMessage() method.
+ @param args.inputManifestPath <required> - Path to input manifest.
+ */
+void call(Map args = [:]) {
+    lib = library(identifier: 'jenkins@main', retriever: legacySCM(scm))
+    def failureMessages = args.failureMessages
+    def passMessages = args.passMessages
```
peterzhuamazon, line 18: Is this the `--continue-on-error` output msg you are checking here?

vars/UpdateBuildFailureIssues.groovy
```diff
@@ -0,0 +1,74 @@
...
+            ghIssueBody = """***Received Error***: **${exactComponentFailureMessage}**.
+                    ${component.name} failed during the distribution build for version: ${currentVersion}.
+                    Please see build log at ${env.RUN_DISPLAY_URL}.
+                    The failed build stage will be marked as unstable(!). Please see ./build.sh step for more details""".stripIndent()
+            createGithubIssue(
+                repoUrl: component.repository,
+                issueTitle: "[AUTOCUT] Distribution Build Failed for ${component.name}-${currentVersion}",
+                issueBody: ghIssueBody,
+                label: "autocut,v${currentVersion}"
+                )
```
peterzhuamazon, line 51: Spacing, also if an issue already opened, does this call just update the description?

vars/UpdateBuildFailureIssues.groovy
```diff
@@ -0,0 +1,74 @@
...
+            sleep(time:3, unit:'SECONDS')
```
peterzhuamazon, line 52: Think you can move the sleep block to the end of the for loop, instead of writing one in `if` and one in `else if`.

vars/UpdateBuildFailureIssues.groovy
```diff
@@ -0,0 +1,74 @@
...
+        } else if (passedComponents.contains(component.name) && !failedComponents.contains(component.name)) {
```
peterzhuamazon, line 53: What is the case for `else` here?

vars/UpdateBuildFailureIssues.groovy
```diff
@@ -0,0 +1,74 @@
...
+ @param Map args = [:] args A map of the following parameters
+ @param args.failureMessages <required> - Failure message retrieved from buildFailureMessage() method.
+ @param args.passMessages <required> - Passing message retrieved from buildFailureMessage() method.
+ @param args.inputManifestPath <required> - Path to input manifest.
+ */
+void call(Map args = [:]) {
+    lib = library(identifier: 'jenkins@main', retriever: legacySCM(scm))
+    def failureMessages = args.failureMessages
+    def passMessages = args.passMessages
+    def yamlFile = readYaml(file: args.inputManifestPath)
```
peterzhuamazon, line 19: Change `yamlFile` to `inputManifestFile` or similar for better understanding.

peterzhuamazon (commented):
Hi @gaiksaya ,

Add some comments. Please review them.

Thanks.

vars/UpdateBuildFailureIssues.groovy
gaiksaya: No it adds comments. I did not change base lib behavior

vars/UpdateBuildFailureIssues.groovy
gaiksaya: Yes these are coming from python code. Again I did not change what was already present in base libraries. Just copied the code.

vars/UpdateBuildFailureIssues.groovy
gaiksaya: No `Failure message retrieved from buildFailureMessage() method.`
Sample: https://github.com/opensearch-project/opensearch-build-libraries/pull/380/files#diff-5bece77137a612237a833245027342b90d0ce60d2b32dd3e4df7cb7287c23806R17-R20

Current implementation: https://github.com/opensearch-project/opensearch-build/blob/main/jenkins/opensearch/distribution-build.jenkinsfile#L888-L898

vars/UpdateBuildFailureIssues.groovy
gaiksaya: Done!

vars/UpdateBuildFailureIssues.groovy
gaiksaya: Nothing. The component will either pass or fail.

vars/UpdateBuildFailureIssues.groovy
gaiksaya: Done!

vars/UpdateBuildFailureIssues.groovy
peterzhuamazon: So then the logic has issues.

If there is anything in failed, it will always go to the `if` block, and ignore `else if` block.

Therefore, you can change the `else if` block to just `else` block?

gaiksaya: No! If the component fail by default it will go in if block and if it passes completely, it will go in else if block. I can add an else block just to print something but the component will either pass or fail. There is no other possible outcome.

peterzhuamazon: After the talk with @gaiksaya offline, this can be improve to be two `if` blocks:

`1st         if (failedComponents.contains(component.name)) {`
will check if the component has a failure. If so, create issue or keep updating comment.

`2nd if (passedComponents.contains(component.name) && !failedComponents.contains(component.name)) {`
will check if component shows up in passes but clear of failed, close issues.

For any other cases such as pass is empty or in pass but also in fail, ignore.

Thanks.

peterzhuamazon (approved):

PR #393, opened 2024-03-07
Add implementation to get component names from Input manifest

Divyaasm:
### Description
Get component names from input manifest 

### Issues Related
https://github.com/opensearch-project/opensearch-build/issues/4005


src/jenkins/InputManifest.groovy
```diff
@@ -53,12 +53,35 @@ class InputManifest {
...
+        String name
+        String repository
+
+        Component(Map data) {
+            this.name = data.name
+            this.repository = data.repository
+        }
+
+    }
+
```
peterzhuamazon, line 76: Any reason not implementation the entire map like here?

https://github.com/opensearch-project/opensearch-build-libraries/blob/14bc0937325533f5541c53c9abfeac80b1e2c7df/src/jenkins/BundleManifest.groovy#L54-L72

Divyaasm: Currently `Components` Property of input Manifest is not used any where, so I have only added arguments which are used by the release-branch creation[ job](https://github.com/opensearch-project/opensearch-build/pull/4520)

peterzhuamazon: I do not agree with that approach.
If input manifest have those properties that you can use, just include them.
Bundle/Build manifest already have those why does input manifest be any different here? Unless it does not support those properties which is another case.

Divyaasm: Sure I can do that. Thanks Peter

peterzhuamazon (approved):

PR #426, opened 2024-05-15
Improve Debian Releases File

marcohald:
### Description
Added Description to aptly repo create
Added  Label and Origin to aply publish snapshot

### Issues Resolved
Closes  https://github.com/opensearch-project/opensearch-build/issues/4485


peterzhuamazon (changes requested):
Hi @marcohald ,

This file is used for both OS and OSD so we cannot hardcode with comments/labels.

Is there any particular benefits of adding them especially when the repo is created with ${jobname} and snapshot with ${jobname}-${repoVersion} already?

Thanks.

marcohald:
@peterzhuamazon The actual generated File https://artifacts.opensearch.org/releases/bundle/opensearch/2.x/apt/dists/stable/InRelease looks like this at the moment
```
-----BEGIN PGP SIGNED MESSAGE-----
Hash: SHA256

Origin: . stable
Label: . stable
Suite: stable
Codename: stable
Date: Tue, 14 May 2024 19:07:45 UTC
Architectures: amd64 arm64
Components: main
Description: Generated by aptly

```
As far as I can see neither ${jobname} or ${repoVersion} is part of that File.
The Problem with that is that it is not as easy to Pin when we cache the Repository with approx.
I could use origin (the hostname of the repo, not the Origin from the Release File) as pin parameter.
The Pinning is to control the Version opensearch that is installed.
In theory there could be multiple Repos on the approx with the same origin that could contain a opensearch package.

As far as i understand the only option that is not from the Release File is the origin to pin.
As found here https://unix.stackexchange.com/a/84812 and here https://manpages.debian.org/bookworm/apt/apt_preferences.5.en.html

peterzhuamazon:
Could you change the comment/origin/label to use different value based on ${jobname} then?
I am happy to have a comment or origin or even label but we shall not hardcode it as this script is used for both OS and OSD and possibly more later on.

Let me know. Thanks.

marcohald:
Thank you for the hint with the different products I didn't thought about that.
I changed it to `${filename}` which should result to "opensearch-dashboards" like the Product: here https://build.ci.opensearch.org/job/distribution-promote-repos/113/console
and "opensearch" for https://build.ci.opensearch.org/job/distribution-promote-repos/111/console
I think the origin could be left hardcoded to opensearch.org what do you think about that?

peterzhuamazon:
I will update your test cases once you commit the requested changes.

Thanks.

marcohald:
Hi @peterzhuamazon 
Sure, thank you for your help with this PR

peterzhuamazon:
Will add test cases in a min.

peterzhuamazon (approved):

peterzhuamazon:
Thanks @marcohald for the contribution and we should be able to see this change take effect in the next release.

peterzhuamazon:
Waiting for https://github.com/opensearch-project/opensearch-build-libraries/pull/427 to bump verision before update the jenkinsfile.

peterzhuamazon:
Use new lib:
* https://github.com/opensearch-project/opensearch-build/pull/4724

PR #436, opened 2024-06-11
Creating Issue Reports for Flaky Test Failures in Gradle Check

prudhvigodithi:
### Description
Gradle Check flaky test failures issue creation

### Issues Resolved
Part of https://github.com/opensearch-project/OpenSearch/issues/13950

### DIFF
--- a/src/gradlecheck/FetchPostMergeFailedTestClass.groovy
+++ b/src/gradlecheck/FetchPostMergeFailedTestClass.groovy
@@ -63,7 +63,7 @@ class FetchPostMergeFailedTestClass {
                         test_class_keyword_agg: [
                                 terms: [
                                         field: "test_class",
-                                        size: 2
+                                        size: 500
                                 ]
                         ]
                 ]
--- a/src/gradlecheck/FetchPostMergeFailedTestClass.groovy
+++ b/src/gradlecheck/FetchPostMergeFailedTestClass.groovy
@@ -1,0 +1,81 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package gradlecheck
+
+import groovy.json.JsonOutput
+import gradlecheck.OpenSearchMetricsQuery
+
+class FetchPostMergeFailedTestClass {
+    String metricsUrl
+    String awsAccessKey
+    String awsSecretKey
+    String awsSessionToken
+    def script
+
+    FetchPostMergeFailedTestClass(String metricsUrl, String awsAccessKey, String awsSecretKey, String awsSessionToken, def script) {
+        this.metricsUrl = metricsUrl
+        this.awsAccessKey = awsAccessKey
+        this.awsSecretKey = awsSecretKey
+        this.awsSessionToken = awsSessionToken
+        this.script = script
+    }
+
+    def getQuery() {
+        def queryMap = [
+                size: 200,
+                query: [
+                        bool: [
+                                must: [
+                                        [
+                                                match: [
+                                                        "invoke_type.keyword": [
+                                                                query: "Post Merge Action",
+                                                                operator: "OR",
+                                                                prefix_length: 0,
+                                                                max_expansions: 50,
+                                                                fuzzy_transpositions: true,
+                                                                lenient: false,
+                                                                zero_terms_query: "NONE",
+                                                                auto_generate_synonyms_phrase_query: true,
+                                                                boost: 1
+                                                        ]
+                                                ]
+                                        ],
+                                        [
+                                                match: [
+                                                        test_status: [
+                                                                query: "FAILED",
+                                                                operator: "OR"
+                                                        ]
+                                                ]
+                                        ]
+                                ]
+                        ]
+                ],
+                aggregations: [
+                        test_class_keyword_agg: [
+                                terms: [
+                                        field: "test_class",
+                                        size: 2
+                                ]
+                        ]
+                ]
+        ]
+
+        def query = JsonOutput.toJson(queryMap)
+        return query.replace('"', '\\"')
+    }
+
+    def getPostMergeFailedTestClass() {
+         def jsonResponse = new OpenSearchMetricsQuery(metricsUrl,awsAccessKey, awsSecretKey, awsSessionToken, script).fetchMetrics(getQuery())
+         def keys = jsonResponse.aggregations.test_class_keyword_agg.buckets.collect { it.key }
+         return keys
+    }
+}
--- a/src/gradlecheck/FetchPostMergeFailedTestName.groovy
+++ b/src/gradlecheck/FetchPostMergeFailedTestName.groovy
@@ -125,6 +125,6 @@ class FetchPostMergeFailedTestName  {
 
     }
     def getPostMergeFailedTestName(testName, gitReference) {
-        return new OpenSearchMetricsQuery(metricsUrl,awsAccessKey, awsSecretKey, awsSessionToken, script).fetchMetrics(getQuery(testName, gitReference))
+        return new OpenSearchMetricsQuery(metricsUrl, awsAccessKey, awsSecretKey, awsSessionToken, script).fetchMetrics(getQuery(testName, gitReference))
     }
 }
--- a/src/gradlecheck/FetchPostMergeFailedTestName.groovy
+++ b/src/gradlecheck/FetchPostMergeFailedTestName.groovy
@@ -1,0 +1,130 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package gradlecheck
+
+import groovy.json.JsonOutput
+import gradlecheck.OpenSearchMetricsQuery
+
+class FetchPostMergeFailedTestName  {
+    String metricsUrl
+    String awsAccessKey
+    String awsSecretKey
+    String awsSessionToken
+    def script
+
+    FetchPostMergeFailedTestName(String metricsUrl, String awsAccessKey, String awsSecretKey, String awsSessionToken, def script) {
+        this.metricsUrl = metricsUrl
+        this.awsAccessKey = awsAccessKey
+        this.awsSecretKey = awsSecretKey
+        this.awsSessionToken = awsSessionToken
+        this.script = script
+    }
+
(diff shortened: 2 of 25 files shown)


prudhvigodithi:
Adding @dblock since a similar setup was done in past for code inside [src/jenkins](https://github.com/opensearch-project/opensearch-build-libraries/tree/main/src/jenkins). I have added a package `gradlecheck` that has the code which can be used by the library `gradleCheckFlakyTestIssueCreate.groovy` and jenkinsfile can simply load and call this library. The only change is I did not  add `implements Serializable` also unlike each class part of [src/jenkins](https://github.com/opensearch-project/opensearch-build-libraries/tree/main/src/jenkins) the code inside `src/gradlecheck` is used only only `gradleCheckFlakyTestIssueCreate` library, I have done this way to make sure the code is organized following OOP methods rather than just dumping all the lines under one library inside `var/gradleCheckFlakyTestIssueCreate`. 

Having this now the unit tests I have used `mockito-core` for testing the code inside `tests/gradlecheck` and works fine (example `./gradlew test --tests=OpenSearchMetricsQueryTest`), but when invoked with existing tests the old tests times out with the following error.

```
2024-06-11T06:44:05.566+0000 [DEBUG] [org.gradle.cache.internal.DefaultFileLockManager] Lock acquired on daemon addresses registry.
2024-06-11T06:44:05.566+0000 [DEBUG] [org.gradle.cache.internal.DefaultFileLockManager] Releasing lock on daemon addresses registry.
2024-06-11T06:44:05.567+0000 [DEBUG] [org.gradle.cache.internal.DefaultFileLockManager] Waiting to acquire shared lock on daemon addresses registry.
2024-06-11T06:44:05.567+0000 [DEBUG] [org.gradle.cache.internal.DefaultFileLockManager] Lock acquired on daemon addresses registry.
2024-06-11T06:44:05.567+0000 [DEBUG] [org.gradle.cache.internal.DefaultFileLockManager] Releasing lock on daemon addresses registry.
```

Just deleting the directory `src/gradlecheck` and running the existing test example `./gradlew clean test --tests=TestUploadTestResults -info -Ppipeline.stack.write=true` again moves forward without any timeout error. 

I'm wondering why just adding one extra package `src/gradlecheck` and testing this with existing test suite is not possible ? @gaiksaya @peterzhuamazon @rishabh6788 @getsaurabh02

prudhvigodithi:
Coming from my above [comment](https://github.com/opensearch-project/opensearch-build-libraries/pull/436#issuecomment-2161461640) the hanging tests issue is due to inheritance used in the `src/gradlecheck` code files, even though this works in runtime and able to complete the end requirement, this hangs during testing. However with the same goal of using `src/gradlecheck` I have modified the code so that its compatible during testing and does not break the existing tests still honoring  fetching the required classes from `src/gradlecheck` for the `gradleCheckFlakyTestChecker` library. 

I have also modified the library `vars/createGithubIssue` that will be used in `gradleCheckFlakyTestChecker` to create/update the issue body with `--body-file` and also allow to edit the issue body (not just comment), this change broke majority of the tests that use `createGithubIssue` directly or indirectly.

prudhvigodithi:
The CI checks are now green and library is tested as jenkins run, sample issues that are created https://github.com/prudhvigodithi/OpenSearch/issues/25 https://github.com/prudhvigodithi/OpenSearch/issues/24.

rishabh6788:
> The CI checks are now green and library is tested as jenkins run, sample issues that are created [prudhvigodithi/OpenSearch#25](https://github.com/prudhvigodithi/OpenSearch/issues/25) [prudhvigodithi/OpenSearch#24](https://github.com/prudhvigodithi/OpenSearch/issues/24).

I see that there are two comments, if I remember correctly wasn't it supposed to be just one comment that would keep getting updated or description?
Also, I feel the way data is being presented can be improved.

prudhvigodithi:
> I see that there are two comments, if I remember correctly wasn't it supposed to be just one comment that would keep getting updated or description?
> Also, I feel the way data is being presented can be improved.

Ya I have removed the `issueEdit` option to test the default behaviour (comment on an open issue rather than editing the body) of the `createGithubIssue` library. I have added it back.

The markdown table is coming from the approval https://github.com/opensearch-project/OpenSearch/issues/13950#issuecomment-2155221675, this has all the information per commit, associated PR, jenkins build and failed tests.
We can always improve based on the feedback once we see this in action by start creating the issues.

vars/createGithubIssue.groovy
```diff
@@ -22,7 +24,7 @@ void call(Map args = [:]) {
     try {
         withCredentials([usernamePassword(credentialsId: 'jenkins-github-bot-token', passwordVariable: 'GITHUB_TOKEN', usernameVariable: 'GITHUB_USER')]) {
             def openIssue = sh(
-                    script: "gh issue list --repo ${args.repoUrl} -S \"${args.issueTitle} in:title\" --label ${label} --json number --jq '.[0].number'",
+                    script: "gh issue list --repo ${args.repoUrl} -S \"${args.issueTitle} in:title\" --label \"${label}\" --json number --jq '.[0].number'",
```
rishabh6788, line 27: In case there are multiple issues returned then is this picking up the top result?

src/gradlecheck/FetchPostMergeFailedTestName.groovy
```diff
@@ -0,0 +1,130 @@
...
+                        ]
+                ]
+        ]
+
+        def query = JsonOutput.toJson(queryMap)
+        return query.replace('"', '\\"')
+
+    }
+    def getPostMergeFailedTestName(testName, gitReference) {
+        return new OpenSearchMetricsQuery(metricsUrl,awsAccessKey, awsSecretKey, awsSessionToken, script).fetchMetrics(getQuery(testName, gitReference))
```
rishabh6788, line 128: nit: missing space after comma.

src/gradlecheck/OpenSearchMetricsQuery.groovy
```diff
@@ -0,0 +1,42 @@
...
+    OpenSearchMetricsQuery(String metricsUrl, String awsAccessKey, String awsSecretKey, String awsSessionToken, def script) {
+        this.metricsUrl = metricsUrl
+        this.awsAccessKey = awsAccessKey
+        this.awsSecretKey = awsSecretKey
+        this.awsSessionToken = awsSessionToken
+        this.script = script
+    }
+
+    def fetchMetrics(String query) {
+        def response = script.sh(
```
rishabh6788, line 30: just for my knowledge, what is 'script.sh'?

rishabh6788:
One thought I have is once the workflow has created an issue, and someone works on it, may be updates the comment or description and strikes-through a few failing tests that have been fixed. 
Now when the workflow runs once again, wouldn't it override the edits made by operator? 
Are we assuming that operator would only be adding new comments on the issue and not editing the comment/description added by with workflow?

vars/createGithubIssue.groovy
prudhvigodithi: Yes Risab that was the idea for this existing library `createGithubIssue`, the logic was with same title for all the AUTOCUT issues, there will be only one issue created for one scenario.

src/gradlecheck/OpenSearchMetricsQuery.groovy
prudhvigodithi: The `script` is a jenkins context and will be injected to the library during runtime. Without the jenkins context the native groovy does have the `sh` inbuit.

prudhvigodithi:
> One thought I have is once the workflow has created an issue, and someone works on it, may be updates the comment or description and strikes-through a few failing tests that have been fixed.
> Now when the workflow runs once again, wouldn't it override the edits made by operator?
> Are we assuming that operator would only be adding new comments on the issue and not editing the comment/description added by with workflow?

AFAIK only maintainers/admins's can edit the issue created by external users and I dont think one can edit the issue created by other user.

src/gradlecheck/FetchPostMergeFailedTestName.groovy
prudhvigodithi: Added thanks

src/gradlecheck/FetchPostMergeFailedTestClass.groovy
```diff
@@ -0,0 +1,81 @@
...
+import gradlecheck.OpenSearchMetricsQuery
+
+class FetchPostMergeFailedTestClass {
+    String metricsUrl
+    String awsAccessKey
+    String awsSecretKey
+    String awsSessionToken
+    def script
+
+    FetchPostMergeFailedTestClass(String metricsUrl, String awsAccessKey, String awsSecretKey, String awsSessionToken, def script) {
```
rishabh6788, line 22: except markdown class, every class has the same constructor and `get_query` methods. Would it be better to have a parent/super class or interface, if supported in groovy class, and have these method defined there and other classes just inherit it and implement in its own way. Method like `getPostMergeFailedTestClass` can have a generic name and implemented in all the sub-classes. 
Not a blocker for me.

prudhvigodithi: That was my initial implementation rishab, take a look at this [commit](https://github.com/opensearch-project/opensearch-build-libraries/commit/b8462b4d890cc2a218bab88924f006d5a1ff0cbd#diff-2087681af514d648a0a6740a4b91375803a761294436f8d010934d4ea7ffdd8d) which has the `extends` and `super` implementation, the jenkins pipeline works but this  `extends` and `super` implementation ran into issues with running the tests (both old and new tests part of the PR) part of this repo, please check my [comment](https://github.com/opensearch-project/opensearch-build-libraries/pull/436#issuecomment-2162026372) here,

src/gradlecheck/OpenSearchMetricsQuery.groovy
prudhvigodithi: Adding more to this, to run a script in groovy the ideal way to use is `.execute()`, but running `.execute()` in jenkins is blocked by the script security plugin, hence the ideal was is to use `sh` given by jenkins and for `sh` to work in libraries it needs jenkins context.

prudhvigodithi:
FYI I have added a comment to an open issue https://github.com/jenkinsci/JenkinsPipelineUnit/issues/595#issuecomment-2164014788 on why we cant test the library directly `gradleCheckFlakyTestChecker` that has the external classes. Here for our case the `gradleCheckFlakyTestChecker` just uses the code part of `src/gradlecheck`, for all the code part of `src/gradlecheck` is tested and covered in `tests/gradlecheck` (Refer the CodeCov https://github.com/opensearch-project/opensearch-build-libraries/pull/436#issuecomment-2161461640).

rishabh6788 (approved):
Approving assuming all search queries are working as expected.

prudhvigodithi:
Thanks @rishabh6788 just pushed a commit change the repo from my fork to `repoUrl: "https://github.com/opensearch-project/OpenSearch",`.

PR #459, opened 2024-07-20
Add `publishDistributionBuildResults` and `publishIntegTestResults` to publish the OpenSearch build and integration test results

prudhvigodithi:
### Description
Add `publishDistributionBuildResults` and `publishIntegTestResults` to publish the OpenSearch build and integration test results

#### Sample Usage
```
publishIntegTestResults(
    version: env.version,
    distributionBuildNumber: env.buildId,
    distributionBuildUrl: "https://build.ci.opensearch.org/blue/organizations/jenkins/distribution-build-opensearch/detail/distribution-build-opensearch/${env.buildId}/pipeline",
    rc: rc,
    rcNumber: env.RC_NUMBER,
    platform: env.platform,
    architecture: env.architecture,
    distribution: env.distribution,
    testReportManifestYml: "test-report.yml"
)
```

```
publishDistributionBuildResults(
    failureMessages: buildMessage(search: 'Error building'),
    passMessages: buildMessage(search: 'Successfully built'),
    inputManifestPath: "manifests/$INPUT_MANIFEST",
    rc: rc,
    rcNumber: env.RC_NUMBER,
    componentCategory: "OpenSearch"
)
```             
### Issues Resolved
Part of https://github.com/opensearch-project/opensearch-metrics/issues/56 and https://github.com/opensearch-project/opensearch-metrics/issues/51

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -127,7 +127,7 @@ jacocoTestReport {
     }
 }
 
-String version = '6.6.0'
+String version = '6.6.2'
 
 task updateVersion {
     doLast {
--- a/vars/publishDistributionBuildResults.groovy
+++ b/vars/publishDistributionBuildResults.groovy
@@ -1,0 +1,171 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+/** Library to fetch the failing Integration test details at the end of Integration Test Jenkins build and index the results to OpenSearch Metrics cluster.
+ *
+ * @param Map args = [:] args A map of the following parameters.
+ * @param args.rc <required> - If the integration tests are running on an RC. 
+ * @param args.rcNumber <required> - The RC number against which the integration test is executed.
+ * @param args.failureMessages <required> - Failure message retrieved from buildFailureMessage() method.
+ * @param args.passMessages <required> - Passing message retrieved from buildFailureMessage() method. Used to get the passed components list.
+ * @param args.distribution <required> - The distribution of the integration test build. Used to get the failed components list.
+ * @param args.componentCategory <required> - The OpenSearch or OpenSearch Dashboards plugin
+ * @param args.inputManifestPath <required> - Path to input manifest.
+ */
+
+import groovy.json.JsonOutput
+import java.text.SimpleDateFormat
+import java.util.Date
+
+void call(Map args = [:]) {
+    def distributionBuildNumber = currentBuild.number
+    def distributionBuildUrl = env.RUN_DISPLAY_URL
+    def buildStartTime = currentBuild.startTimeInMillis
+    def currentDate = new Date()
+    def formattedDate = new SimpleDateFormat("MM-yyyy").format(currentDate)
+    def rc = args.rc
+    def rcNumber = args.rcNumber.toInteger()
+    def indexName = "opensearch-distribution-build-results-${formattedDate}"
+    def failureMessages = args.failureMessages
+    def passMessages = args.passMessages
+    def componentCategory = args.componentCategory
+    def inputManifest = readYaml(file: args.inputManifestPath)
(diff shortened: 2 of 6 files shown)


build.gradle
```diff
@@ -127,7 +127,7 @@ jacocoTestReport {
     }
 }
 
-String version = '6.6.1'
+String version = '6.6.2'
```
peterzhuamazon, line 130: Should be at least a minor bump.

vars/publishIntegTestResults.groovy
```diff
@@ -0,0 +1,226 @@
...
+                    curl -XPOST "${METRICS_HOST_URL}/$indexName/_bulk" --aws-sigv4 \"aws:amz:us-east-1:es\" --user \"${awsAccessKey}:${awsSecretKey}\" -H \"x-amz-security-token:${awsSessionToken}\" -H "Content-Type: application/x-ndjson" --data-binary "@${testRecordsFile}"
+                else
+                    echo "File Does not exist. No tests records to process."
+                fi
+        """
+        }
+    }
+}
+
+def generateJson(component, version, integTestBuildNumber, integTestBuildUrl, distributionBuildNumber, distributionBuildUrl, buildStartTime, rc, rcNumber, platform, architecture, distribution, componentCategory, componentResult, testReportManifestYmlUrl, withSecurity, withSecurityYml, withSecurityStdout, withSecurityStderr, withoutSecurity, withoutSecurityYml, withoutSecurityStdout, withoutSecurityStderr) {
```
peterzhuamazon, line 199: Possible to combine both generateJson block to a common lib?
Or maybe rename to a different func name to avoid confusion?

Thanks.

prudhvigodithi: Thank you. The `generateJson` method generates different `key:value` paired documents for each use case, such as for distribution builds and integration tests. Therefore, combining them into a common method would be overkill, as the purpose of `generateJson` is unique for each use case. As for the naming, I kept `generateJson` constant for both `publishDistributionBuildResults` and `publishIntegTestResults` because the purpose is the same and it is easily understood by developers. The `generateJson` should be an interface method but not possible with groovy jenkins libraries, here is some discussion from past https://github.com/opensearch-project/opensearch-build-libraries/pull/436#discussion_r1637169462.

peterzhuamazon: Got it, thanks.

peterzhuamazon (approved):

build.gradle
prudhvigodithi: Sure @peterzhuamazon, adding @gaiksaya for some inputs on the minor bump.

prudhvigodithi: Looking back again I agree since we added two new libs we should go with `6.7.0`.

vars/publishDistributionBuildResults.groovy
```diff
@@ -0,0 +1,170 @@
...
+                            },
+                            "component_category": {
+                                "type": "keyword"
+                            },
+                            "component_build_result": {
+                                "type": "keyword"
+                            }
+                        }
+                    }
+                }'
```
gaiksaya, line 117: Should we define this JSON in groovy using JSON parses? I believe shell wont be able to catch the typos

prudhvigodithi: Initially I have tried that Sayali, but end of the day its the index mapping using during `curl` with `-d`, with groovy the curl is rejecting the data parsing with format exception, hence moved the block to shell in a proper json format and then adding with via `-d` had no parsing issues. Added some good tests with normalized data `normalizeString(calledCommands[0]) == normalizeString(expectedCommandBlock)`, if the index mapping is changed lets says with typos the tests will fail and and as well OpenSearch itself will fail to index once the index mapping is created.

gaiksaya (approved):
Are we okay with lib naming? OR should we be more specific where we are publishing the results to? (metrics cluster in this case)

prudhvigodithi:
> Are we okay with lib naming? OR should we be more specific where we are publishing the results to? (metrics cluster in this case)

The README has the details Sayali, else the library name is becoming very long.

PR #497, opened 2024-09-13
Add script to terminate stale jobs

rishabh6788:
### Description
Add script to terminate stale jobs. 
This will help in checking and aborting redundant builds for jobs. 

There are no tests due to lack of mocking support for Jenkins instance methods used in the script. 

### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/5008 

### DIFF
--- a/vars/abortStaleJenkinsJobs.groovy
+++ b/vars/abortStaleJenkinsJobs.groovy
@@ -1,0 +1,63 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+/** Library to fetch failing tests at the end of gradle-check run and index the results in an OpenSearch cluster.
+ *
+ * @param Map args = [:] args A map of the following parameters
+ * @param args.jobName <required> - The name of the jenkins job.
+ * @param args.lookupTime <optional> - Fetch builds from past N hours for the job, defaults to 6 hours.
+ */
+
+import java.time.Instant
+import java.time.temporal.ChronoUnit
+import jenkins.model.Jenkins
+import hudson.model.Result
+
+void call(Map args = [:]) {
+    String jobName = args.jobName.toString()
+    long lookupTime = isNullOrEmpty(args.lookupTime.toString()) ? 6 : Long.parseLong(args.lookupTime.toString())
+
+    if (isNullOrEmpty(jobName)) {
+        throw new IllegalArgumentException("Error: jobName is null or empty")
+    }
+
+    def currentBuildNumber = currentBuild.number
+    def currentBuildDescription = currentBuild.description
+    def endTime = Instant.now()
+    def startTime = endTime.minus(lookupTime, ChronoUnit.HOURS)
+    def startMillis = startTime.toEpochMilli()
+    def endMillis = endTime.toEpochMilli()
+
+    // Add sleep to let job-id get assigned to queued jobs when triggered via generic webhook url
+    sleep(15)
+
+    def currentJob = Jenkins.instance.getItemByFullName(jobName)
+
+    //Fetch all builds for the job based on look up time provided
+    def builds = currentJob.getBuilds().byTimestamp(startMillis,endMillis)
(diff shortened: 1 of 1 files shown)


vars/abortStaleJenkinsJobs.groovy
```diff
@@ -0,0 +1,63 @@
...
+
+    // Add sleep to let job-id get assigned to queued jobs when triggered via generic webhook url
+    sleep(15)
+
+    def currentJob = Jenkins.instance.getItemByFullName(jobName)
+
+    //Fetch all builds for the job based on look up time provided
+    def builds = currentJob.getBuilds().byTimestamp(startMillis,endMillis)
+    for (build in builds) {
+        if (build.isBuilding() && currentBuildNumber > build.number && currentBuildDescription == build.description) {
```
gaiksaya, line 45: Are we terminating purely based on build description?

vars/abortStaleJenkinsJobs.groovy
```diff
@@ -0,0 +1,63 @@
...
+            try {
+                build.doStop()
+                println "Aborted build #${build.number} for ${build.description}"
+            }
+            catch (Exception e) {
+                if (build.result == Result.ABORTED) {
```
gaiksaya, line 51: Do we need to add any checks for failed/already stopped builds?

vars/abortStaleJenkinsJobs.groovy
```diff
@@ -0,0 +1,63 @@
...
+                    println "Build #${build.number} was already aborted, possibly by someone else"
```
gaiksaya, line 52: nit: Might be jenkins agents issue as well. Keeping it generic
```suggestion
                    println "Build #${build.number} is already aborted!"
```

vars/abortStaleJenkinsJobs.groovy
peterzhuamazon: Same question.

vars/abortStaleJenkinsJobs.groovy
```diff
@@ -0,0 +1,63 @@
...
+    sleep(15)
+
+    def currentJob = Jenkins.instance.getItemByFullName(jobName)
+
+    //Fetch all builds for the job based on look up time provided
+    def builds = currentJob.getBuilds().byTimestamp(startMillis,endMillis)
+    for (build in builds) {
+        if (build.isBuilding() && currentBuildNumber > build.number && currentBuildDescription == build.description) {
+            try {
+                build.doStop()
```
peterzhuamazon, line 47: Is doStop enough for our case as these might take long to stop.
Do we want to immediately stop it since the agents are deleted anyway?

vars/abortStaleJenkinsJobs.groovy
rishabh6788: AFAIK when a gradle-check is triggered on a pull request the build description remains the same for each new build of the job. Let me know if I can have a check on any other attribute?

vars/abortStaleJenkinsJobs.groovy
rishabh6788: the `build.isBuilding()` in if condition only iterates over in-progress builds. This try/catch is to catch edge condition where two jobs are executing this block at the same time and then one might fail due to job already terminated by other one. We don't want to fail the build job if this block fails to process due to race condition. 
This block just catches this condition without stopping the flow.

vars/abortStaleJenkinsJobs.groovy
rishabh6788: I think this is a graceful stop which also takes care of stopping the underlying running process. 
If we use `doKill` then it will hard stop the pipeline and there is no guarantee that underlying gradle process would be stopped on the host. 
If the agents are deleted for aborted or killed pipelines then yes we can use `doKill` instead of stop. Let me know.

peterzhuamazon: Yes each agent will only run one build, then delete after.
Tho probably the existing graceful stop is enough for the time being.

peterzhuamazon (approved):
Thanks @rishabh6788 . We will use this method for now and please update the version.

We can think about later if there is a way for github actions to send out a call to Jenkins api to cancel, before they themselves get canceled than using an additional script to clean up jenkins.

Thanks.

PR #567, opened 2024-11-11
Add ci-group arg for OSD integ tests

rishabh6788:
### Description
Add ci-group arg for OSD integ tests. 

### Issues Resolved


vars/processAndRunOsdIntegTest.groovy
```diff
@@ -0,0 +1,54 @@
+/*
```
gaiksaya, line 1: Not understanding why we need to add this library separately? What is the difference?

vars/processAndRunOsdIntegTest.groovy
```diff
@@ -0,0 +1,54 @@
...
+    }
+    else {
+        echo "Not on Windows, unstash repository+artifacts"
+    }
+
+    sh("rm -rf test-results")
+
+    runIntegTestScript(
+            jobName: "${BUILD_JOB_NAME}",
+            componentName: args.localComponent,
```
gaiksaya, line 47: The lib name says OSD so I think componentName will be static here to `OpenSearch-Dashboards`? Maybe add a check and fail if that's not the case?

vars/runIntegTestScript.groovy
```diff
@@ -55,13 +55,16 @@ void call(Map args = [:]) {
     String switchCommandStart = switchUser.equals('true') ? "su `id -un 1000` -c \"env PATH=\$PATH $javaHomeCommand" : "env PATH=\$PATH $javaHomeCommand"
     String switchCommandEnd = switchUser.equals('true') ? '"' : ''
 
+    String testManifest = "manifests/${args.testManifest}"
```
gaiksaya, line 58: This would be a breaking change for the lib right? Looking at https://github.com/opensearch-project/opensearch-build/blob/main/jenkins/opensearch-dashboards/integ-test.jenkinsfile#L286?
Maybe bump the major version.

tests/jenkins/jobs/ProcessAndRunOsdIntegTest_Jenkinsfile
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+pipeline {
+    agent none
+    stages {
+        stage('integ-test') {
+            steps {
+                script {
+                    processAndRunOsdIntegTest(
+                        localComponent: 'OpenSearch',
```
gaiksaya, line 17: This should fail right? Since the lib as the name suggests only tests OSD?

vars/processAndRunOsdIntegTest.groovy
rishabh6788: It moves logic here https://github.com/opensearch-project/opensearch-build/blob/main/jenkins/opensearch-dashboards/integ-test.jenkinsfile#L251-L290 to avoid duplicating the same code in if-else block.

vars/processAndRunOsdIntegTest.groovy
rishabh6788: No, for now OSD will be passed but tomorrow if any other component decides to use ci-groups, this need not change, all will be controlled from calling njenkinsfile.

vars/runIntegTestScript.groovy
rishabh6788: Don't think it will be breaking, we are just moving the path details from jenkinsfile to script. 
I am okay to do a major version bump.

tests/jenkins/jobs/ProcessAndRunOsdIntegTest_Jenkinsfile
rishabh6788: This is just a dummy component name to check component without any ci-group.

vars/processAndRunOsdIntegTest.groovy
```diff
@@ -0,0 +1,54 @@
...
+                downloadPath: "${artifactPath}/",
+                bucketName: "${ARTIFACT_BUCKET_NAME}",
+                localPath: "${WORKSPACE}/artifacts",
+                force: true
+        )
+        sh("cp -a ${WORKSPACE}/artifacts/${artifactPath} ${WORKSPACE}")
+        sh("rm -rf ${WORKSPACE}/artifacts")
+    }
+    else {
+        echo "Not on Windows, unstash repository+artifacts"
```
gaiksaya, line 40: Should this be
```suggestion
        echo "Not on Windows, skipping unstashing repository+artifacts"
```
I dont see any action being taken in else statement. Just the echo.

gaiksaya:
Can you also please add library definitions to these 2 libraries. Something similar to https://github.com/opensearch-project/opensearch-build-libraries/blob/main/vars/publishDistributionBuildResults.groovy#L10-L19
Looks like we missed a lot of them for old ones. Might be a good idea to use this opportunity to fix these 2 for now.

vars/runIntegTestScript.groovy
rishabh6788: Done

tests/jenkins/jobs/ProcessAndRunOsdIntegTest_Jenkinsfile
rishabh6788: Done

vars/processAndRunOsdIntegTest.groovy
rishabh6788: Not sure.

rishabh6788: If it is not a blocker we can revisit this later.

vars/runIntegTestScriptForOSD.groovy
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    sh("rm -rf test-results")
+
+    runIntegTestScript(
+            jobName: "${BUILD_JOB_NAME}",
+            componentName: args.localComponent,
+            buildManifest: "${BUILD_MANIFEST}",
+            testManifest: "${TEST_MANIFEST}",
+            localPath: "${WORKSPACE}/${distribution}",
```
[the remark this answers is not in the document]
gaiksaya: This sounds risky if someone needs to use the lib directly. It wont know what to specify. Those args are Parameters in the workflow thats why we can use it as ENV variable. I believe it we need to add as args and add checks.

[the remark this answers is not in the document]
rishabh6788: Done.

PR #611, opened 2025-03-12
Add qualifier support to metrics queries

gaiksaya:
### Description
Add qualifier support to metrics queries

### Issues Resolved
Work around solution for https://github.com/opensearch-project/opensearch-build/issues/5386


src/jenkins/ComponentIntegTestStatus.groovy
```diff
@@ -106,6 +115,13 @@ class ComponentIntegTestStatus {
                         ]
                 ]
         ]
+        if (this.qualifier != null && this.qualifier != "None") {
```
peterzhuamazon, line 118: Too many of this calls can we replace with: https://github.com/opensearch-project/opensearch-build-libraries/blob/a0e5039b16a98ebb1347d83c4a14dc53491cec14/vars/abortStaleJenkinsJobs.groovy#L63

src/jenkins/ReleaseCandidateStatus.groovy
```diff
@@ -71,43 +73,19 @@ class ReleaseCandidateStatus {
...
-                _source: "distribution_build_number",
-                sort: [
-                    [
-                        distribution_build_number: [
-                            order: "desc"
-                        ]
+            queryMap.sort[0].remove("rc_number")
+            queryMap.query.bool.filter.add([
+                    match_phrase: [
+                            rc_number: "${rcNumber}"
```
peterzhuamazon, line 79: What is this trying to do?

vars/checkRequestAssignReleaseOwners.groovy
```diff
@@ -12,19 +12,20 @@ import java.time.LocalDate
...
     String action = args.action ?: 'check'
-    def version = extractVersionFromManifest(inputManifest[0])
     def now = LocalDate.now()
     def monthYear = String.format("%02d-%d", now.monthValue, now.year)
     def maintainersIndex = "maintainer-inactivity-${monthYear}"
 
     // Parameter validation
     validateParameters(args)
+    def manifestYaml = readYaml(file: inputManifest[0])
+    def version = manifestYaml.build.version
```
peterzhuamazon, line 28: Input Manifest would have qualifier field do you not need to extract it?

src/jenkins/ComponentIntegTestStatus.groovy
gaiksaya: We want to check for "None" as that is what is indexed by default for non-qualifiers while indexing data.

src/jenkins/ReleaseCandidateStatus.groovy
gaiksaya: Refactored the code to easy logic. See the diff. It does the same. If rc_number is passed the query is different than the default one

vars/checkRequestAssignReleaseOwners.groovy
gaiksaya: Not required. The queries used here deal with github issues and nothing to do with build or test data

peterzhuamazon: But isnt that we want to also make sure issues follow qualifier?

src/jenkins/ReleaseCandidateStatus.groovy
peterzhuamazon: I guess I didnt get why you need to `            queryMap.sort[0].remove("rc_number")`.

vars/checkRequestAssignReleaseOwners.groovy
gaiksaya: No labels and issues do not have qualifiers.
Both class objects here: https://github.com/opensearch-project/opensearch-build-libraries/blob/main/vars/checkRequestAssignReleaseOwners.groovy#L42-L43 do not need qualifiers in their queries.

src/jenkins/ReleaseCandidateStatus.groovy
gaiksaya: Looks like diff is bit messed up. Basically instead of reassigning the `queryMap` here: https://github.com/opensearch-project/opensearch-build-libraries/blob/main/src/jenkins/ReleaseCandidateStatus.groovy#L73-L111
just using a good logic to add and remove the required fields before it

peterzhuamazon: Got it now, you are just removing from sort and adding in query.

Thanks.

src/jenkins/ComponentIntegTestStatus.groovy
gaiksaya: Added a private method to incorporate above logic.
Thanks!

peterzhuamazon (approved):

PR #626, opened 2025-03-18
Add library to check and notify release notes based on old workflow

gaiksaya:
### Description
The python [release workflow](https://github.com/opensearch-project/opensearch-build/tree/main/src/release_notes_workflow) in build-repo generates a markdown table in the format: https://github.com/opensearch-project/opensearch-build/issues/3747#issuecomment-2704366641
This change adds below things:
- Adds a class to parse the markdown file and return each component with respective values mapped to the keys.
- Adds a new library that will `check` or `notify` the release owners about the missing release notes. 

### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/5332

### DIFF
--- a/resources/release/criteria-not-met-template.md
+++ b/resources/release/criteria-not-met-template.md
@@ -1,4 +1,4 @@
-Hi ${RELEASE_OWNER}, </br>
+Hi @${RELEASE_OWNER}, </br>
 
 The below ${TYPE_OF_CRITERIA} criteria for your component has not been met. </br>
 Please review the issue and address it with high priority. </br>
--- a/resources/release/missing-release-notes.md
+++ b/resources/release/missing-release-notes.md
@@ -1,0 +1,5 @@
+Hi, </br>
+
+This component is missing release notes at ${BRANCH} ref. Please add them on priority in order to meet the entrance criteria for the release.
+
+Thank you!
--- a/resources/release/missing-release-notes.md
+++ b/resources/release/missing-release-notes.md
@@ -1,5 +1,6 @@
 Hi, </br>
 
-This component is missing release notes at ${BRANCH} ref. Please add them on priority in order to meet the entrance criteria for the release.
+This component is missing release notes at ${BRANCH} ref. Please add them on priority in order to meet the entrance criteria for the release. </br>
+Please check out the [guidelines](https://github.com/opensearch-project/opensearch-plugins/blob/main/RELEASE_NOTES.md) for the release notes. </br>
 
 Thank you!
--- a/vars/checkReleaseNotes.groovy
+++ b/vars/checkReleaseNotes.groovy
@@ -1,0 +1,118 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+import jenkins.ReleaseMetricsData
+import jenkins.ParseReleaseNotesMarkdownTable
+/**
+ * Library to check and notify missing release notes.
+ * @param Map args = [:] args A map of the following parameters
+ * @param args.version <required> - Release version. eg: 3.0.0
+ * @param args.dataTable <required> - Markdown data table file in the format: https://github.com/opensearch-project/opensearch-build/issues/3747#issuecomment-2704366641 eg: ./table.md
+ * @param args.action <optional> - Action to perform. Default is 'check'. Acceptable values are 'check' and 'notify'.
+ */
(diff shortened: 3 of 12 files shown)


vars/checkReleaseNotes.groovy
```diff
@@ -0,0 +1,118 @@
...
+ * @param args.dataTable <required> - Markdown data table file in the format: https://github.com/opensearch-project/opensearch-build/issues/3747#issuecomment-2704366641 eg: ./table.md
+ * @param args.action <optional> - Action to perform. Default is 'check'. Acceptable values are 'check' and 'notify'.
+ */
+void call(Map args = [:]) {
+    String action = args.action ?: 'check'
+    // Parameter check
+    validateParameters(args, action)
+    def version = args.version.tokenize('-')[0]
+
+    def parsedContent = new ParseReleaseNotesMarkdownTable(readFile(args.dataTable)).parseReleaseNotesMarkdownTableRows()
```
zelinh, line 24: Do we pass the complete table here or we use the link?

resources/release/missing-release-notes.md
zelinh, line 1: I think we could also link a release notes guidance to this issue template.

resources/release/criteria-not-met-template.md
```diff
@@ -1,4 +1,4 @@
-Hi ${RELEASE_OWNER}, </br>
+Hi @${RELEASE_OWNER}, </br>
 
 The below ${TYPE_OF_CRITERIA} criteria for your component has not been met. </br>
```
zelinh, line 3: Is it possible that we can tell what exact criteria that's not met here?

zelinh (commented):
Thanks @gaiksaya, could you look into the CodeCov failing here?

vars/checkReleaseNotes.groovy
gaiksaya: Entire table as a file.

resources/release/criteria-not-met-template.md
gaiksaya: We have not decided yet. I just added `@` as it was missed.

resources/release/missing-release-notes.md
gaiksaya: Added the guidelines. Thanks! Also the component release issue already had the guidelines but yeah its good to have.

gaiksaya:
> Thanks @gaiksaya, could you look into the CodeCov failing here?

Added a test for exception as well.

vars/checkReleaseNotes.groovy
```diff
@@ -0,0 +1,118 @@
...
+        withCredentials([
+                string(credentialsId: 'jenkins-health-metrics-account-number', variable: 'METRICS_HOST_ACCOUNT'),
+                string(credentialsId: 'jenkins-health-metrics-cluster-endpoint', variable: 'METRICS_HOST_URL')]) {
+            withAWS(role: 'OpenSearchJenkinsAccessRole', roleAccount: "${METRICS_HOST_ACCOUNT}", duration: 900, roleSessionName: 'jenkins-session') {
+                def metricsUrl = env.METRICS_HOST_URL
+                def awsAccessKey = env.AWS_ACCESS_KEY_ID
+                def awsSecretKey = env.AWS_SECRET_ACCESS_KEY
+                def awsSessionToken = env.AWS_SESSION_TOKEN
+
+                ReleaseMetricsData releaseMetricsData = new ReleaseMetricsData(metricsUrl, awsAccessKey, awsSecretKey, awsSessionToken, version, this)
```
zelinh, line 72: These credential usages are universal for all different plugins right? Any harm we extract them outside the loop to avoid multiple times calling?
e.g. 
```
withCredentials([...] {
   withAWS(...) {
       ...
       ReleaseMetricsData releaseMetricsData = new ReleaseMetricsData(...)
       componentsWithFalseStatus.each { component ->
            ...
        }
   }
}

Comment (gaiksaya) [reply vars/checkReleaseNotes.groovy:72]: They are universal but passed down in each call. If the withCrerdentials are only available within the block AFAIK. And to make those calls `getReleaseIssue` we need those credentials..

Comment (zelinh) [reply vars/checkReleaseNotes.groovy:72]: Yeah. Since the entire function relies on credentials, I'm thinking we could put all related operations within a single credential block. This would prevent redundant calls for each component. Would it be possible to do it this way?

Comment (gaiksaya) [reply vars/checkReleaseNotes.groovy:72]: We should not. Only use credentials where required. In case we want to add more functionality to component with false status, that should go outside credential block but within loop.
For example:
```
    componentsWithFalseStatus.each { component ->
        withCredentials([
                ........
                ........
                    addComment(releaseIssueUrl, ghCommentContent)
                }
         // do something else with these components here that has nothing to do with metrics credentials
            }
        }

zelinh (approved):

PR #680, opened 2025-05-16
Create new build library to ingest smoke test report to metrics cluster

zelinh:
### Description
Create new build library to ingest smoke test report to metrics cluster. 

### Issues Resolved
Part of https://github.com/opensearch-project/opensearch-build/issues/5320

### DIFF
--- a/tests/jenkins/TestPublishSmokeTestResults.groovy
+++ b/tests/jenkins/TestPublishSmokeTestResults.groovy
@@ -1,0 +1,257 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package jenkins.tests
+
+import groovy.json.JsonSlurper
+import org.junit.Before
+import org.junit.Test
+
+class TestPublishSmokeTestResults extends BuildPipelineTest {
+
+    @Override
+    @Before
+    void setUp() {
+        super.setUp()
+        binding.setVariable('currentBuild', [
+            number: 123,
+            startTimeInMillis: System.currentTimeMillis(),
+        ])
+        binding.setVariable('env', [
+            RUN_DISPLAY_URL: 'http://example.com/build/123'
+        ])
+        binding.setVariable('sh', { cmd -> println cmd })
+        binding.setVariable('readFile', { filePath -> 'components:\n- name: component1\n  configs:\n    - name: GET___cat_plugins\n      status: PASS\n    - name: POST___bulk\n      status: FAIL' })
+        binding.setVariable('writeFile', { params -> println params.text })
+        binding.setVariable('withCredentials', { creds, closure -> closure() })
+        helper.registerAllowedMethod("withAWS", [Map, Closure], { args, closure ->
+            closure.delegate = delegate
+            return helper.callClosure(closure)
+        })
+        binding.setVariable('curl', { params -> println params })
+    }
+
+    @Test
+    void testIndexSmokeTestData() {
+        def indexName = 'test-index'
+        def testRecordsFile = 'test-records.ndjson'
+
+        def script = loadScript('vars/publishSmokeTestResults.groovy')
+
+        def calledCommands = new ArrayList()
+        script.metaClass.sh = { String command ->
+            calledCommands << command
+            if (command.contains("curl -I")) {
+                return "HTTP/1.1 200 OK"
+            } else if (command.contains("curl -s -XPUT") && command.contains("test-index")) {
+                return '{"acknowledged":true}'
+            } else if (command.contains("curl -XPOST") && command.contains("test-index")) {
+                return '{"took":10, "errors":false}'
+            } else {
+                throw new IllegalArgumentException("Unexpected command: $command")
+            }
+        }
+
+        script.indexSmokeTestData(indexName, testRecordsFile)
+
+        def expectedCommandBlock = '''set +e
+        set +x
+        echo "INDEX NAME IS test-index"
+                INDEX_MAPPING='{
+                    "mappings": {
+                        "properties": {
+                            "architecture": {
+                                "type": "keyword"
+                            },
+                            "build_start_time": {
+                                "format": "epoch_millis",
+                                "type": "date"
+                            },
+                            "component": {
+                                "type": "keyword"
+                            },
+                            "component_category": {
+                                "type": "keyword"
+                            },
+                            "component_repo": {
+                                "type": "keyword"
+                            },
+                            "component_repo_url": {
+                                "type": "keyword"
+                            },
+                            "distribution": {
+                                "type": "keyword"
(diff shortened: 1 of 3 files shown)


prudhvigodithi:
Thanks @zelinh,overall the code LGTM but is it possible to show the sample visualizations ? Example https://github.com/opensearch-project/OpenSearch/issues/11217#issuecomment-2140815607.

Once merged and indexed, want to understand how does the indexed docs look like and what are the plans to visualize this.

zelinh:
> Thanks @zelinh,overall the code LGTM but is it possible to show the sample visualizations ? Example [opensearch-project/OpenSearch#11217 (comment)](https://github.com/opensearch-project/OpenSearch/issues/11217#issuecomment-2140815607).
> 
> Once merged and indexed, want to understand how does the indexed docs look like and what are the plans to visualize this.

Screenshot 2025-05-20 at 12 10 56 PM

Hey Prudhvi here is how the visualization will look like. We can filter by test results to only look into test failures. In this sample, the component yaml is not available because of the relative path for generating test report in our beta Jenkins is different from prod Jenkins.

prudhvigodithi:
Thanks, @gaiksaya can you also take a look as well. Thanks

vars/publishSmokeTestResults.groovy
```diff
@@ -0,0 +1,256 @@
...
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+/** Library to fetch the failing Integration test details at the end of Integration Test Jenkins build and index the results to OpenSearch Metrics cluster.
+ *
+ * @param Map args = [:] args A map of the following parameters.
+ * @param args.distributionBuildUrl <required> - The jenkins distribution build number.
+ * @param args.testReportManifestYml <required> - The generated test report YAML file using test report workflow.
+ * @param args.jobName <required> - The smoke test job name, used in `testReportManifestYmlUrl`.
```
gaiksaya, line 15: Why do we need jobName as required? Seeing that is being only used to get `testReportManifestYmlUrl`
Would it make sense to directly set arg `testReportManifestYmlUrl` instead? 
Also what is the difference between `testReportManifestYml` and `testReportManifestYmlUrl`

vars/publishSmokeTestResults.groovy
```diff
@@ -0,0 +1,256 @@
...
+                            }
+                        }
+                    }
+                }'
+                curl -I "${METRICS_HOST_URL}/${indexName}" --aws-sigv4 \"aws:amz:us-east-1:es\" --user \"${awsAccessKey}:${awsSecretKey}\" -H \"x-amz-security-token:${awsSessionToken}\" | grep -E "HTTP\\/[0-9]+(\\.[0-9]+)? 200"
+                if [ \$? -eq 0 ]; then
+                    echo "Index already exists. Indexing Results"
+                else
+                    echo "Index does not exist. Creating..."
+                    create_index_response=\$(curl -s -XPUT "${METRICS_HOST_URL}/${indexName}" --aws-sigv4 \"aws:amz:us-east-1:es\" --user \"${awsAccessKey}:${awsSecretKey}\" -H \"x-amz-security-token:${awsSessionToken}\" -H 'Content-Type: application/json' -d "\${INDEX_MAPPING}")
```
gaiksaya, line 188: Maybe in future we can use a top level util package for this instead of curl commands with right initialization and checks? Wondering if it is possible to use java-client too.
@prudhvigodithi

vars/publishSmokeTestResults.groovy
```diff
@@ -0,0 +1,256 @@
...
+        architecture: architecture,
+        distribution: distribution,
+        component_category: componentCategory,
+        test_report_manifest_yml: testReportManifestYmlUrl,
+        component_test_spec_name: componentTestSpecName,
+        component_test_spec_yml: componentTestSpecYml,
+        component_test_spec_result: componentTestSpecResult,
+        component_test_spec_cluster_stdout: componentTestSpecClusterStdout,
+        component_test_spec_cluster_stderr: componentTestSpecClusterStderr,
+        "component_test_spec_test_stdout": componentTestSpecTestStdout,
```
gaiksaya, line 252: Are `"` expected?

vars/publishSmokeTestResults.groovy
zelinh: `testReportManifestYml` is the test report file which will be read through to retrieve test results.
`testReportManifestYmlUrl` is the URL which will be indexed into metrics cluster as a link to the report.

vars/publishSmokeTestResults.groovy
```diff
@@ -0,0 +1,256 @@
...
+
+void call(Map args = [:]) {
+
+    def isNullOrEmpty = { str ->
+        str == null || (str instanceof String && str.trim().isEmpty())
+    }
+    // Check if any args is equals to null or it is a test run
+    if (isNullOrEmpty(args.distributionBuildUrl) || isNullOrEmpty(args.testReportManifestYml) || isNullOrEmpty(args.jobName) || args.jobName.equals('dummy_job')) {
+        return null
+    }
```
gaiksaya, line 30: Maybe separate into a different method such as https://github.com/opensearch-project/opensearch-build-libraries/blob/main/vars/checkRequestAssignReleaseOwners.groovy#L66-L82 that would throw right error about what is missing. Also would help adding tests for the same.

tests/jenkins/TestPublishSmokeTestResults.groovy
```diff
@@ -0,0 +1,257 @@
...
+        def args = [
+            distributionBuildUrl: "http://example.com/distribution/456",
+            testReportManifestYml: "path/to/testReportManifest.yml",
+            jobName: ""
+        ]
+
+        def result = script.call(args)
+
+        assert result == null
+    }
```
gaiksaya, line 252: These should throw an error right? Should hard fail the job.

gaiksaya (commented):
Overall looks good. Can you please handle the errors correctly. Example, for missing args fail had using `error` rather than just returning null.

vars/publishSmokeTestResults.groovy
zelinh: Would we want the entire job to fail if the indexing step doesn't complete successfully?

From my perspective, this library serves as an add-on to the existing smoke tests and is primarily responsible for ingesting results into the metrics cluster. If the ingestion fails, it shouldn’t impact the core test execution or its results.

This seems consistent with how we handle indexing in the integration tests as well, as shown here:
https://github.com/opensearch-project/opensearch-build-libraries/blob/91a1c453f4e14731d164de2e0feb4d4273b28487/vars/publishIntegTestResults.groovy#L25-L32

gaiksaya: I see what you mean. Do we have guard rails in place that wrong results are not being indexed? Such as empty values or nulls? 
Reason being if we are taking after actions based on those data, we need to make sure we have the correct data.

gaiksaya (approved):
Lets get this in and then iterate based on the data that is indexed. We can try throwing an exception here for failures and the workflow using this library can handle the exceptions as the way they want.

PR #725, opened 2025-07-14
Update query to retrieve entries for timer for gradlecheck test failure issue creation

Divyaasm:
### Description
Iclude Timer along with Pull Request in InvokeType while creating github issues in OpenSearch Repo.
### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/5537

### DIFF
--- a/src/gradlecheck/CreateMarkDownTable.groovy
+++ b/src/gradlecheck/CreateMarkDownTable.groovy
@@ -21,8 +21,17 @@ class CreateMarkDownTable {
     }
 
     def createMarkdownTable() {
+        def groupRows = this.tableData.groupBy{it.gitReference }
+        def tableRows = groupRows.collect { Ref, rows ->
+            def pullRequestLink = rows.collect { it.pullRequestLink }.unique().join('<br><br>')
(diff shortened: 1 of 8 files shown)


prudhvigodithi:
LGTM, @Divyaasm can you also confirm if the `CreateMarkDownTable` does not create a new for different runs of `main`. Today here is the sample issue https://github.com/opensearch-project/OpenSearch/issues/18695 which has new row for each commit, it would be good if `main` has one row with all the details.

Divyaasm:
> LGTM, @Divyaasm can you also confirm if the `CreateMarkDownTable` does not create a new row for different runs of `main`. Today here is the sample issue [opensearch-project/OpenSearch#18695](https://github.com/opensearch-project/OpenSearch/issues/18695) which has new row for each commit, it would be good if `main` has one row with all the details.

Yes @prudhvigodithi, it lists multiple entries matching with main in the same row without repetition of rows with ref main for each different value.

prudhvigodithi (approved):

Divyaasm:
Hey @prudhvigodithi , also added code to merge multiple rows with common gitreference. This should not create any duplicate rows now.

PR #742, opened 2025-07-30
Add gradle check issue creation only for main branch

Divyaasm:
### Description
Adds gradle check issue creation only for main branch
Fixes build link error in issue body.

### Issues Resolved

### DIFF
--- a/src/gradlecheck/CreateMarkDownTable.groovy
+++ b/src/gradlecheck/CreateMarkDownTable.groovy
@@ -21,7 +21,7 @@ class CreateMarkDownTable {
     }
 
     def createMarkdownTable() {
-        def groupRows = this.tableData.groupBy{it.gitReference }
+        def groupRows = this.tableData.findAll { it.gitReference && !["2.19", "2.x", "1.x"].contains(it.gitReference)}.groupBy{ it.gitReference }
         def tableRows = groupRows.collect { Ref, rows ->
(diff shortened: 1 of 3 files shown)


src/gradlecheck/CreateMarkDownTable.groovy
```diff
@@ -21,7 +21,7 @@ class CreateMarkDownTable {
     }
 
     def createMarkdownTable() {
-        def groupRows = this.tableData.groupBy{it.gitReference }
+        def groupRows = this.tableData.findAll { it.gitReference && !["2.19", "2.x", "1.x"].contains(it.gitReference)}.groupBy{ it.gitReference }
```
prudhvigodithi, line 24: Instead can we have just look up for `main` and ignore rest, this is to avoid adding multiple versions or .x branches in future.

prudhvigodithi: I mean it should be possible right when you gitReference what it contains and move forward if it contains main.

prudhvigodithi: I got it now. So we create everything for main and commitsID's. So this code works for all except "2.19", "2.x", "1.x" ?

Divyaasm: Yes Prudhvi. I will filter only main and commitid's on the issue body

PR #822, opened 2026-05-05
Fix incremental build for windows

gaiksaya:
### Description
Fix incremental build for windows

### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/5839

### DIFF
--- a/vars/buildManifest.groovy
+++ b/vars/buildManifest.groovy
@@ -70,7 +70,7 @@ void call(Map args = [:]) {
                 ).trim()
             } else {
                 DISTRIBUTION_BUILD_NUMBER = bat(
-                        script:  "curl -sL ${PUBLIC_ARTIFACT_URL}/${DISTRIBUTION_JOB_NAME}/${revision}/index/${DISTRIBUTION_PLATFORM}/${DISTRIBUTION_ARCHITECTURE}/${distribution}/index.json | jq -r \".latest\"",
+                        script:  "@curl -sL ${PUBLIC_ARTIFACT_URL}/${DISTRIBUTION_JOB_NAME}/${revision}/index/${DISTRIBUTION_PLATFORM}/${DISTRIBUTION_ARCHITECTURE}/${distribution}/index.json | jq -r \".latest\"",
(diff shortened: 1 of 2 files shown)


vars/buildManifest.groovy
```diff
@@ -70,7 +70,7 @@ void call(Map args = [:]) {
                 ).trim()
             } else {
                 DISTRIBUTION_BUILD_NUMBER = bat(
-                        script:  "curl -sL ${PUBLIC_ARTIFACT_URL}/${DISTRIBUTION_JOB_NAME}/${revision}/index/${DISTRIBUTION_PLATFORM}/${DISTRIBUTION_ARCHITECTURE}/${distribution}/index.json | jq -r \".latest\"",
+                        script:  "@curl -sL ${PUBLIC_ARTIFACT_URL}/${DISTRIBUTION_JOB_NAME}/${revision}/index/${DISTRIBUTION_PLATFORM}/${DISTRIBUTION_ARCHITECTURE}/${distribution}/index.json | jq -r \".latest\"",
```
peterzhuamazon, line 73: I think use curl.exe is better than using @ to bypass alias in batch files.
If you can test that before merging it would be better.
Thanks.

gaiksaya: The actual issue @ solves: when Jenkins bat(returnStdout: true) runs a command, it captures everything printed to stdout including the echoed command prompt (C:\...>curl ...). The @ suppresses that echo. Using curl.exe instead of curl doesn't help with that — you'd still get the prompt echoed.

Windows has been broken long back and negates all build time. Even if linux is building incrementally, because of windows it does not matter as build takes usual time.

We can try this out, monitor for sometime and then take the call.

peterzhuamazon: Have confirmed with Sayali that this is just same as `@echo off` so we will just approve this.
Was thinking of replacing with pwsh but this is good enough for now.

Thanks.

peterzhuamazon (approved):
Version bump if possible, Thanks for this.

PR #841, opened 2026-06-11
Refactor standalone release workflows to use prerelease and add notifications

gaiksaya:
### Description
Since bot access has been removed for all repositories, the release workflows are unable to read the draft release. This PR fixes that. It now triggers only for pre-release. Would need additional changes in all repositories to switch from draft release creation to pre-release release. 
Also adds notification mechanism to both slack and GitHub
Maintainers now need to manually publish the pre-release to actual release on GitHub as bot does not have access anymore. 

### Issues Resolved

### DIFF
--- a/vars/publishNotification.groovy
+++ b/vars/publishNotification.groovy
@@ -7,15 +7,18 @@
  * compatible open source license.
  */
 void call(Map args = [:]) {
-    text = ([
+    def parts = [
         "${args.icon}",
-        "JOB_NAME=${JOB_NAME}",
-        "BUILD_NUMBER=[${BUILD_NUMBER}]",
+        "JOB_NAME=${env.JOB_NAME}",
+        "BUILD_NUMBER=[${env.BUILD_NUMBER}]",
         "MESSAGE=${args.message}",
-        "BUILD_URL: ${BUILD_URL}",
-        "MANIFEST: ${args.manifest}",
-        args.extra
-    ] - null).join("\n")
+        "BUILD_URL: ${env.BUILD_URL}",
+    ]
+    if (args.manifest) {
+        parts.add("MANIFEST: ${args.manifest}")
+    }
+    parts.add(args.extra)
+    text = (parts - null).join("\n")
 
     def secret_webhook = [
         [envVar: 'WEBHOOK_URL', secretRef: "op://opensearch-release-secrets/webhook/${args.credentialsId}"]
--- a/vars/standardReleasePipelineWithGenericTrigger.groovy
+++ b/vars/standardReleasePipelineWithGenericTrigger.groovy
@@ -7,7 +7,7 @@
  * compatible open source license.
  */
 
-/** A standard release pipeline for OpenSearch projects including generic triggers. A tag or a draft release can be used as a trigger using this library. The defaults are all set to trigger via a draft release. If the release is successful, the release can be published by using right params.
+/** A standard release pipeline for OpenSearch projects including generic triggers. A tag or a pre-release can be used as a trigger using this library. The defaults are all set to trigger via a pre-release. If the release is successful, a GitHub issue is created for maintainers to publish the release.
 @param Map arguments = [:] arguments A map of the following parameters
 @param body <Required> - A closure containing release steps to be executed in release stage.
 @param arguments.tokenIdCredential <Required> - Credential id containing token for trigger authentication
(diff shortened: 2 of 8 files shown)


vars/publishNotification.groovy
```diff
@@ -7,15 +7,18 @@
...
         "BUILD_URL: ${BUILD_URL}",
-        "MANIFEST: ${args.manifest}",
-        args.extra
-    ] - null).join("\n")
+    ]
+    if (args.manifest) {
+        parts.add("MANIFEST: ${args.manifest}")
+    }
+    parts.add(args.extra)
+    text = (parts - null).join("\n")
```
peterzhuamazon, line 21: Any reason you need to change to this?

vars/standardReleasePipelineWithGenericTrigger.groovy
```diff
@@ -47,16 +47,17 @@ void call(Map arguments = [:], Closure body) {
                     [key: 'ref', value: (arguments.jsonValue ?: '$.release.tag_name')],
                     [key: 'repository', value: '$.repository.html_url'],
                     [key: 'action', value: '$.action'],
-                    [key: 'isDraft', value: '$.release.draft'],
+                    [key: 'isPreRelease', value: '$.release.prerelease'],
```
peterzhuamazon, line 49: Note if it is pre-release then artifact must be attached when the release is creating.
After created, you cannot update the attached artifacts anymore due to immutable release.

vars/standardReleasePipelineWithGenericTrigger.groovy
```diff
@@ -73,21 +74,28 @@ void call(Map arguments = [:], Closure body) {
...
-                                sh "curl -J -L -H 'Accept: application/octet-stream' -H 'Authorization: Bearer ${GITHUB_TOKEN}' ${assetUrl} -o artifacts.tar.gz && tar -xvf artifacts.tar.gz"
+                                }
+
+                                if (assetUrl == null) {
+                                    def availableAssets = parsedJson.collect { it.name }.join(', ')
+                                    error("Asset '${assetName}' not found in release assets. Available assets: [${availableAssets}]. Assets URL: ${assets_url}")
+                                }
+
+                                echo "Downloading artifacts from ${assetUrl}"
+                                sh "curl -s --fail -J -L -H 'Accept: application/octet-stream' -H 'Authorization: Bearer ${GITHUB_TOKEN}' ${assetUrl} -o artifacts.tar.gz && tar -xvf artifacts.tar.gz"
```
peterzhuamazon, line 97: This is just a call out that a year long incorrect setup here that still runs.
https://github.com/opensearch-project/opensearch-protobufs/blob/main/.github/workflows/publish-release.yml#L40
`tar -C build -cvf "opensearch-protobufs-java.tar.gz` will create a tar file, not tgz file.
But since you use
`tar -xvf artifacts.tar.gz` to extract it still runs here, but technically this is not a tgz without `-z`.

I dont think we can change anything at this point but just called out here.

vars/standardReleasePipelineWithGenericTrigger.groovy
```diff
@@ -105,14 +113,37 @@ void call(Map arguments = [:], Closure body) {
             success {
                 script {
                     if (arguments.publishRelease && release_url != null) {
-                        withSecrets(secrets: secret_github_bot){
-                            sh "curl -X PATCH -H 'Accept: application/vnd.github+json' -H 'Authorization: Bearer ${GITHUB_TOKEN}' ${release_url} -d '{\"tag_name\":\"${tag}\",\"draft\":false,\"prerelease\":false}'"
+                        withSecrets(secrets: secret_github_bot) {
+                            def codeOwners = sh(
+                                script: "gh api repos/${repository.replaceAll('https://github.com/', '')}/contents/.github/CODEOWNERS --jq '.content' 2>/dev/null | tr -d '\\n' | base64 -d | grep -oP '@[\\w/-]+' | sort -u | tr '\\n' ' ' || echo ''",
```
peterzhuamazon: Introduce GH instead of using git directly?
gh would need gh auth login right? And the token can be retrieved in plain text.
If this can be done in git api call directly with curl then why not.

vars/publishNotification.groovy
gaiksaya: Yes this was pretty static to build level of things. Parts is optional making it more env agnostic if someone wants to add more details to the notification messages

vars/standardReleasePipelineWithGenericTrigger.groovy
gaiksaya: Yes nothing changes. All workflows needs attach it while creating

vars/standardReleasePipelineWithGenericTrigger.groovy
gaiksaya: That would multiple steps to get codeowners file. Looks like this is better way. Or issue creation below this we are using gh

peterzhuamazon: How are you authenticating with gh then?

gaiksaya: Looks like it is using the GH TOKEN tested it out here: https://github.com/gaiksaya/spring-data-opensearch/issues/16

peterzhuamazon (approved):

PR #862, opened 2026-07-17
Support cluster write operations and add base for release state docs

gaiksaya:
### Description
Introduces validated document models (ReleaseSchedule, ReleaseCriterion, ReleaseDecision) for the release state machine, each enforcing required and enum-like fields so malformed docs fail fast. 
Also adds `indexDocument` write method to OpenSearchMetricsQuery

### Issues Resolved
Part of https://github.com/opensearch-project/oscar-ai-bot/issues/169

### DIFF
--- a/src/jenkins/ReleaseDecision.groovy
+++ b/src/jenkins/ReleaseDecision.groovy
@@ -1,0 +1,81 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package jenkins
+
+/**
+ * A Go/No-Go decision record for the opensearch_release_state index.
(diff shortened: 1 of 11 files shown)


src/jenkins/ReleaseDecision.groovy
```diff
@@ -0,0 +1,81 @@
...
+        return value
+    }
+
+    private static String optionalOneOf(Map args, String key, List<String> allowed) {
+        String value = args[key]
+        if (value != null && !allowed.contains(value)) {
+            throw new IllegalArgumentException("ReleaseDecision: '${key}' must be one of ${allowed}, got '${value}'.")
+        }
+        return value
+    }
```
peterzhuamazon: Can these be saved to a class? I think you add them to nearly all groovy files.

peterzhuamazon (approved):

src/jenkins/ReleaseDecision.groovy
gaiksaya: Good catch. I abstracted it under utils. And used only for new code base. We can reuse for other classes later is required.
Thanks!

PR #888, opened 2026-08-04
Add a library to check unpatched vulnerabilities

gaiksaya:
### Description
Adding a library to get data from security advisories cluster. This change will be incorporated in `indexReleaseState` to check CVE exit criteria.

### Issues Resolved
https://github.com/opensearch-project/oscar-ai-bot/issues/169

### DIFF
--- a/src/jenkins/SecurityAdvisoryData.groovy
+++ b/src/jenkins/SecurityAdvisoryData.groovy
@@ -1,0 +1,198 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package jenkins
+
+import groovy.json.JsonOutput
+import utils.SecurityAdvisoriesQuery
+
+/**
+ * Reads unpatched-vulnerability data from the security advisories OpenSearch cluster.
+ *
+ * Mirrors ReleaseStateData in shape (a thin, testable wrapper over a query client), but reads a
+ * different, read-only cluster. The data supports the release exit criterion "no unpatched
+ * vulnerabilities of medium or higher severity that have been publicly known for more than 60 days":
+ *
+ *   1. resolveVersionTag  - map a release version to the branch tag the scans are keyed on
+ *                           (e.g. 3.8.0 / 3.8 -> origin/3.8, main -> origin/main). Never the exact
+ *                           3.8.0 tag, which only exists after release.
+ *   2. getLatestScansIndex - resolve the newest concrete scans-NNNNNN index.
+ *   3. getOpenVulnerabilityIds - the open (non-excluded) CVE ids scanned for that branch tag.
+ *   4. getAgedMediumOrHigherAdvisories - of those ids, the ones the advisories index says are
+ *                           medium/high/critical AND were published on or before the cutoff.
+ *
+ * Every method returns concrete data or throws; callers (checkUnpatchedVulnerabilities) translate a
+ * throw into an 'unknown' criterion so a failed lookup never reads as a clean release.
+ */
+class SecurityAdvisoryData {
+
+    /** Severities that count toward the criterion (medium or higher). */
+    static final List<String> BLOCKING_SEVERITIES = ['CRITICAL', 'HIGH', 'MEDIUM']
+
+    /** The advisories index (an alias pointing at the current advisories index). */
+    static final String ADVISORIES_INDEX = 'advisories'
+
+    /**
+     * Max CVE ids per advisories terms lookup. OpenSearch's default max_terms_count is 65536; a
+     * conservative batch keeps payloads small and mirrors the security_advisories agent.
+     */
+    private static final int ADVISORIES_BATCH_SIZE = 1000
+
+    /** Page size for scan/advisory searches, matching the agent's default query size. */
+    private static final int QUERY_SIZE = 1000
+
+    def script
+    SecurityAdvisoriesQuery advisoriesQuery
+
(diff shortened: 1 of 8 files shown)


vars/checkUnpatchedVulnerabilities.groovy
```diff
@@ -0,0 +1,108 @@
...
+ *         silently passing.
+ */
+Map<String, List<String>> call(Map args = [:]) {
+    if (!args.version) {
+        error('version parameter is required.')
+    }
+
+    def secret_advisories_cluster = [
+        [envVar: 'ADVISORIES_HOST_ACCOUNT', secretRef: 'op://opensearch-release-secrets/aws-accounts/security-advisories-account-number'],
+        [envVar: 'ADVISORIES_HOST_URL', secretRef: 'op://opensearch-release-secrets/security-advisories-cluster/security-advisories-cluster-endpoint']
```
peterzhuamazon, line 46: These two exist?

gaiksaya: No would need to create them. Once PR is approved I can reach out to you before merging.

src/jenkins/SecurityAdvisoryData.groovy
```diff
@@ -0,0 +1,198 @@
...
+     * @return map of project name -> sorted list of its open CVE ids
+     */
+    Map<String, List<String>> getOpenVulnerabilitiesByProject(String scansIndex, String branchTag) {
+        // project.name is carried through so results can be scoped to release components once that
+        // filter lands upstream (security-advisories#132).
+        def query = shellEscape([
+            size    : QUERY_SIZE,
+            sort    : [['timestamp.scan': [order: 'desc']]],
+            collapse: [field: 'project.name'],
+            _source : ['project.name', 'vulnerabilities.id', 'vulnerabilities.excluded'],
```
shreyah963: In advisories, the scanner stores both id and aliases for vulnerabilities. In the website we're currently using vulnerabilities.aliases which contains all identifiers(mostly both CVE and GHSA). Using alias might make it more resilient and match the website approach.

tests/jenkins/TestCheckUnpatchedVulnerabilities.groovy
```diff
@@ -0,0 +1,142 @@
...
+                        {"id": "CVE-1", "excluded": false},
+                        {"id": "CVE-2", "excluded": true}
+                      ]
+                    }
+                  },
+                  {
+                    "_source": {
+                      "project": {"name": "SQL"},
+                      "vulnerabilities": [
+                        {"id": "CVE-3", "excluded": false},
```
shreyah963, line 64: We also have excluded-advisories index and the website's neglected page checks that index to review exemptions. This still would work but if someone adds an exemption in the website after the last scan runs this check will still flag that CVE as blocking until the next scan cycle. If we want it to reflect as soon as an exemption is added we could go through excluded-advisories index for real-time exemptions.

src/jenkins/SecurityAdvisoryData.groovy
```diff
@@ -0,0 +1,198 @@
...
+     * @param scansIndex the concrete scans index (from getLatestScansIndex)
+     * @param branchTag  the resolved project.tag (from resolveVersionTag)
+     * @return map of project name -> sorted list of its open CVE ids
+     */
+    Map<String, List<String>> getOpenVulnerabilitiesByProject(String scansIndex, String branchTag) {
+        // project.name is carried through so results can be scoped to release components once that
+        // filter lands upstream (security-advisories#132).
+        def query = shellEscape([
+            size    : QUERY_SIZE,
+            sort    : [['timestamp.scan': [order: 'desc']]],
```
shreyah963: The website also sorts by timestamp.commit desc to ensure it picks the scan for the latest code state where timestamp.commit is primary sort and timestamp.scan is tiebreaker since sometimes project can be re-scanned against an older commit hash. Maybe we could also use the same approach here.
```
sort: [
                { 'timestamp.commit': { order: 'desc' } },
                { 'timestamp.scan': { order: 'desc' } },
              ],
```

src/utils/SecurityAdvisoriesQuery.groovy
```diff
@@ -0,0 +1,97 @@
...
+        this.script.println("Advisories index '${targetIndex}' returned ${response?.hits?.hits?.size() ?: 0} hit(s).")
+        return response
+    }
+
+    /**
+     * Runs a cluster-wide search (GET /_search with no index prefix), used to resolve which
+     * concrete scan index is the most recent across all indices.
+     * @param query the SigV4-shell-escaped query body
+     */
+    def searchAllIndices(String query) {
```
shreyah963, line 59: Could we use ${advisoriesUrl}/scans-*/_search to restrict to only scan indices and avoid other shard queries?

gaiksaya: This is a helper utility so needs to be flexible. We are also querying `advisories`. In future if more come up it can be used for that as well

src/jenkins/SecurityAdvisoryData.groovy
gaiksaya: Addressed

src/jenkins/SecurityAdvisoryData.groovy
gaiksaya: added vulnerabilities.aliases to _source and now collect every identifier (id + all aliases) per vulnerability.

tests/jenkins/TestCheckUnpatchedVulnerabilities.groovy
gaiksaya: Added new method getExemptedAliasesByProject() reads the exemption index live so an exemption added after the last scan is honored immediately.

shreyah963 (approved):

peterzhuamazon (approved):

PR #913, opened 2026-08-15
Add lib to update release issue with criteria checks

gaiksaya:
### Description
Adds a library to update the release issue with most updated chore checks from metrics cluster as well as github APIs. 
The manual exit and entrance criteria are left untouched.
Also catalogues all the release criteria in one place

### Issues Resolved
 https://github.com/opensearch-project/oscar-ai-bot/issues/169


vars/updateReleaseIssue.groovy
```diff
@@ -0,0 +1,137 @@
...
+ * Rewrites the status circle of each chore-verified criterion row to reflect its indexed status.
+ * A row is matched by locating a chore criterion's keyword in the Criteria cell; only its status
+ * cell (the first circle on the line) is replaced, leaving the other columns untouched. Rows whose
+ * status is unknown, or has no circle mapping, are left as-is so a transient failure never blanks a
+ * circle the release manager can see.
+ */
+private String applyChoreCircles(String issueBody, Map<String, String> statuses) {
+    def choreByKeyword = ReleaseCriterionCatalog.choreCriteria()
+    // Split after each newline so line terminators are preserved and an unchanged body round-trips.
+    return issueBody.split(/(?<=\n)/).collect { line ->
```
peterzhuamazon: Isnt this is a problem where you search every table in this issue?
If later someone add another table and has similar terms it will be error out?

src/jenkins/ReleaseStateData.groovy
```diff
@@ -108,6 +108,48 @@ class ReleaseStateData {
...
+                    filter: [
+                        [term: [doc_type: 'criterion']],
+                        [term: [version: version]],
+                        [term: [source: 'chore_check']]
+                    ]
+                ]
+            ]
+        ]
+        return JsonOutput.toJson(queryMap).replace('"', '\\"')
+    }
```
peterzhuamazon: if say, both OS and OSD has all_integration_tests_passing, how can the query distinguish between OS and OSD, because both have all_integration_tests_passing, and your query does not specify which one to pull the data. Is it possible you pulled OSD data and apply to OS?

vars/updateReleaseIssue.groovy
```diff
@@ -0,0 +1,137 @@
...
+ */
+void call(Map args = [:]) {
+    String version = args.version
+    String releaseIssue = args.releaseIssue
+    String action = args.action ?: 'update_criteria'
+    if (!version || !releaseIssue) {
+        error('version and releaseIssue are required.')
+    }
+
+    def issueNumber = (releaseIssue =~ /^https:\/\/github\.com\/opensearch-project\/opensearch-build\/issues\/(\d+)$/)
```
peterzhuamazon: This review is entirely from AI:

------

Do not retain the regex Matcher in the Pipeline method

issueNumber is a java.util.regex.Matcher, not the extracted issue number. Jenkins Pipeline can pause at steps such as withSecrets and sh so it can resume after a controller restart; when it pauses, Jenkins saves the local variables still held by call(). A Matcher cannot be saved, so the build can fail even though issueRef was already extracted.

For example, the job can reach withSecrets, the controller restarts, and Jenkins then fails while restoring issueNumber instead of continuing to edit the issue. The unit test passes because it runs synchronously and does not exercise a Jenkins pause/resume.

Please parse the URL in a helper that returns only a String, leaving no Matcher in call():

``` groovy
@NonCPS
private String extractIssueRef(String releaseIssue) {
    def matcher = releaseIssue =~ /^https:\/\/github\.com\/opensearch-project\/opensearch-build\/issues\/(\d+)$/
    return matcher.matches() ? matcher.group(1) : null
}
```

Then use String issueRef = extractIssueRef(releaseIssue) and keep the existing invalid-URL check.

src/jenkins/ReleaseStateData.groovy
gaiksaya: Thank you for catching that. It has been addressed in the recent commit. The data is now collected per product category for 2 different exit criteria.

vars/updateReleaseIssue.groovy
gaiksaya: Added the non-cps method.

vars/updateReleaseIssue.groovy
gaiksaya: Refactored this to load specific table based on headings (###)

src/jenkins/ReleaseStateData.groovy
```diff
@@ -110,34 +110,39 @@ class ReleaseStateData {
...
-        Map<String, String> statusByCriterion = [:]
+        Map<String, Map<String, String>> statusByCriterion = [:]
         hits.each { hit ->
             def doc = hit._source
-            if (doc?.criterion_name && doc?.status) {
-                statusByCriterion[doc.criterion_name] = doc.status
+            if (doc?.criterion_name && doc?.product && doc?.status) {
+                Map<String, String> byProduct = statusByCriterion.get(doc.criterion_name, [:])
+                // Hits are sorted newest first, so the first status seen per product is the latest.
+                if (!byProduct.containsKey(doc.product)) {
```
peterzhuamazon: Need to confirm if we use product key or release_type key.

peterzhuamazon (approved):

src/jenkins/ReleaseStateData.groovy
gaiksaya: release_type is exclusively for advisories only which is being used here https://github.com/opensearch-project/opensearch-build-libraries/pull/913/changes#diff-f036a8a24dca6e0f7f6c6b36b1f01d7de0c049ff50b354ab82b2ffa2573fa6f7R145
This query is for per product exit criteria

